PR #1, opened 2019-03-28
Fix Yaml formatting

abhi-g:
OWNERS file is not functional without the Yaml fix.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/1)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,3 +1,3 @@
 approvers:
-  neuromage
-  zhenghuiwang
+  - neuromage
+  - zhenghuiwang


neuromage:
@abhi-g should we add yourself and @jlewi as owners here too? 
/lgtm
/approve

abhi-g:
If we end up making PRs to this repo, we'll request add.

abhi-g:
Also, I'll need to force merge this PR, as owners file is not functional in current form

PR #2, opened 2019-04-04
Roadmap for v0.6

zhenghuiwang:
List the plan for v0.6 release. Issues will be created for these items.

/cc @neuromage @jlewi @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/2)

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -2,10 +2,10 @@
 
 1. CUJ.The CUJ is to track model development in Notebook and Kubeflow Pipeline. ([deck](http://bit.ly/kf-meta-cuj))
 
-2. API and metadata specifications. The goal is to define API and a set of built-in metadata specifications that can support the CUJ above; The API should be able to handle customized metadata specifications.
+2. API and metadata specifications. The goal is to define API and a set of built-in metadata specifications that can support the CUJ above; The API should be able to handle customized metadata specifications.(issues [#5](https://github.com/kubeflow/metadata/issues/5) [#8](https://github.com/kubeflow/metadata/issues/8))
 
-3. Implementations. We plan to leverage [MLMD](https://github.com/google/ml-metadata) for an implementation of the API, because it is already used by Kubeflow Pipeline. Different implementations are welcome and should be easily discoverable.
+3. Implementations. We plan to leverage [MLMD](https://github.com/google/ml-metadata) for an implementation of the API, because it is already used by Kubeflow Pipeline. Different implementations are welcome and should be easily discoverable.(issues [#4](https://github.com/kubeflow/metadata/issues/4) [#9](https://github.com/kubeflow/metadata/issues/9))
 
-4. Python library. We plan to create a Python library for logging metadata in Notebook and Kubeflow Pipeline.
+4. Python library. We plan to create a Python library for logging metadata in Notebook and Kubeflow Pipeline.(issue [#6](https://github.com/kubeflow/metadata/issues/6))
 
-5. User interface. We plan to have a page in Kubeflow dashboard to show models and workflows in Notebook and Kubeflow Pipeline.
+5. User interface. We plan to have a page in Kubeflow dashboard to show models and workflows in Notebook and Kubeflow Pipeline. (issue [#7](https://github.com/kubeflow/metadata/issues/7))
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,10 +1,10 @@
 # Roadmap for v0.6
 
-1. CUJ.The CUJ is to track model development in Notebook and Kubeflow Pipeline. [(deck)](http://bit.ly/kf-meta-cuj)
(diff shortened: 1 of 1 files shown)


neuromage:
/lgtm

Thanks for writing this up!

jlewi:
/hold
Lets give Manasi and the other folks that chimed in on slack a chance to chime in.

roadmap.md
```diff
@@ -0,0 +1,11 @@
+# Roadmap for v0.6
+
+1. CUJ.The CUJ is to track model development in Notebook and Kubeflow Pipeline. ([deck](http://bit.ly/kf-meta-cuj))
+
+2. API and metadata specifications. The goal is to define API and a set of built-in metadata specifications that can support the CUJ above; The API should be able to handle customized metadata specifications.
+
+3. Implementations. We plan to leverage [MLMD](https://github.com/google/ml-metadata) for an implementation of the API, because it is already used by Kubeflow Pipeline. Different implementations are welcome and should be easily discoverable.
+
+4. Python library. We plan to create a Python library for logging metadata in Notebook and Kubeflow Pipeline.
```
terrytangyuan: Is the logged metadata mutable? Can the mutability be controlled by the user? If so, we may need all CRUD for the logged metadata.

zhenghuiwang: Yes, in general metadata is mutable and CRUD API will be provided. But some field are immutable, such as the uuid, type.

terrytangyuan: Sounds good.

mvartakAtVerta:
@zhenghuiwang thanks for putting this out there. 

2 process things:
- Could you please cc folks on the PR in the future? thanks to @jlewi for asking to put on slack, helps keep everyone in the loop. 
- For things that require consensus, can you give people a deadline to chime in? (i unfortunately missed this bc of a conference, but would liked to next time)

One major concern:
- When/how was the decision made to use MLMD? Don't think that was discussed -- last time we spoke (i.e., 4 days ago), it seemed like that was not the preferred option. This is the kind of thing that should be discussed in the community

zhenghuiwang:
> * For things that require consensus, can you give people a deadline to chime in? (i unfortunately missed this bc of a conference, but would liked to next time)

That's a good suggestion. I should have explicitly stated that: I plan to go over it during next community meeting call. This PR won't land before that.

> * When/how was the decision made to use MLMD? Don't think that was discussed -- last time we spoke (i.e., 4 days ago), it seemed like that was not the preferred option. This is the kind of thing that should be discussed in the community

It is not a decision, still open to discussion. Let's  use this PR as the forum.

The major advantages are 
1) MLMD plans to support multiple types of DBs. (it supports MySQL, but plan to support other types). We don't need to do that ourselves.
2) KF Pipeline already has some experience using it for managing pipeline metadata. 
3) It is an open sourced by TFX team and may have smother integration with other TFX tools. 

Having said all that, MLMD is just an implementation. We shouldn't use MLMD API as KF's but define our own API for our CUJs.

What other systems do you think we should consider for an implementation of our API?

jlewi:
@mpvartak It would be good to clarify what issues need consensus and along those lines minimize the number of issues that need consensus.

With respect to implementation the ROADMAP explicitly calls out the fact that there will be multiple implementations

> Implementations. We plan to leverage MLMD for an implementation of the API, because it is already used by Kubeflow Pipeline. Different implementations are welcome and should be easily discoverable.(issues #8 #9)

How about we start filing issues for the different implementations people want to work on and we leave it to individuals to pick which implementations they want to contribute to?

> Could you please cc folks on the PR in the future? thanks to @jlewi for asking to put on slack, helps keep everyone in the loop. 

We will certainly make a good faith effort to call people's attention to important issues and give people a chance to weigh in.

Just to set expectations though this is not strictly speaking a requirement per [Kubeflow contributing guidelines](https://www.kubeflow.org/docs/about/contributing/).

If folks would like to be notified about issues and PRs some suggestions

   * Subscribe to GitHub notifications
   * Add yourself as a reviewer in the OWNERs files  (doesn't guarantee that all PRs will be assigned to you).

The ROADMAP is a living document so we can add more to it in the future. I'd encourage folks to open up follow on PRs adding items that they are willing to commit to and opening up issues related to that.

My interpretation of the current ROADMAP is that this is what @zhenghuiwang and @neuromage are looking to commit to in 0.6

ywskycn:
> Having said all that, MLMD is just an implementation. We shouldn't use MLMD API as KF's but define our own API for our CUJs.

@zhenghuiwang , doea it mean KF's APIs wraps MLMD APIs for now, but also can support others?

zhenghuiwang:
> @zhenghuiwang , doea it mean KF's APIs wraps MLMD APIs for now, but also can support others?

It should be a superset of what MLMD offers in terms of functionalities. The API may look very different but KF's API uses MLMD to model ML workflows, which is what MLMD is built for.

In addition to that, KF metadata store needs to handle data about organizing workflows, such as workspace for users, HP experiment spec, and etc. Also we need better schema registration for customized metadata.

PR #3, opened 2019-04-07
Add basic gRPC-based server with HTTP endpoints (using grpc-gateway)

neuromage:
This PR sets up the basic scaffolding for the Metadata API server.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/3)


README.md
```diff
@@ -1,2 +1,20 @@
...
+To update BUILD file rules, run:
+```
+bazel run //:gazelle
+```
+
+To build and test everything using the `go` command:
+```
+go build ./...
+go test ./...
+```
```
zhenghuiwang: Can you add the instruction on how to run a server locally?

api/generate_proto.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# !/bin/sh
+
+# This file generates API sources from the protocol buffers defined in this
+# directory using Bazel, then copies them back into the source tree so they can
+# be checked-in.
+
+set -e
+set -x
+BAZEL_BINDIR=$(bazel info bazel-bin)
```
zhenghuiwang: Why do we use Bazel here? Can we use standard Go module to build?

If we use Go toolchain, there is no need to autogen Bazel build targets.

README.md
neuromage: Done.

api/generate_proto.sh
neuromage: We need `bazel` to build the proto files. This ensures our build is hermetic, and not dependent on any developer's machine or the tools installed therein. Instead, all protos always get built using `bazel`s toolchain, which will be pinned to the versions specified in the `WORKSPACE` file. 

Note that `go build` for building the server continues to work as before though.

Issue #4, opened 2019-04-08
Define a set of metadata specifications for v0.6


zhenghuiwang:
These metadata specifications should

1. cover metadata needed for [v0.6 CUJ](http://bit.ly/kf-meta-cuj), including model, hyperparameters, metrics etc.
2. show how customized metadata can be defined in a similar way.

These metadata specifications are preloaded during server start, while customized ones are registered via endpoints.

zhenghuiwang:
Some thoughts on what format/tool should be used to define the spec.

We need to handle customized types during runtime. Protobuf is not good at this case, because it needs to compile and generate code. [JSON schema](https://json-schema.org/) seems to be a good fit:

1. Schema can be referenced and extended.
2. Many libraries are available to compile schema and validate JSON during runtime.

Following this [JSON schema guideline on defining complex schema](https://json-schema.org/understanding-json-schema/structuring.html), it would be good if we can extract common fields and reuse them. It is also important to categorize metadata by semantic meaning.

Here is my attempt to define a type hierarchy of metadata of ML workflows. A child is a subtype of its parent. For example, we have a generic `entity` ([schema draft](https://gist.github.com/zhenghuiwang/c3b8f526be3af298876f61b4b5fd737c)). `artifact`([schema draft](https://gist.github.com/zhenghuiwang/9fbb721c723998a8167ef51bb02a2ee1)) is a subtype of `entity`; `data_set`([schema draft](https://gist.github.com/zhenghuiwang/edf1170878eaff75341c4063ddb44c7c)) is a subtype of `artifact`.

```
entity
├── artifact
│       ├── data_set
│       └── model
│
├── executable
│
└── container
         ├── workspace
         └── katib_experiment
```

`entity` is a piece of metadata to represent an entity in Kubeflow. There are three categories of entities:
`artifact` represents input data or derived data in ML workflow, such as data set and model.
`executable` represents an executable in ML workflow, such as training code and data transformation.
`container` represents a group of artifacts, executables and other containers. E.g. workspace and experiments.

terrytangyuan:
+1 for using JSON schema. Two questions:

1. Should we use “data_set” or “dataset” (personally prefer “dataset”)? 
2. Could you clarify what kind of data transformation is executable? In some scenarios, data transformations may be part of the model itself, e.g. a TensorFlow graph containing both preprocessing and the trained model. There are also other types of  model artifacts such as PMML and PFA where people might also store preprocessing logic within the model.

[...]

PR #10, opened 2019-04-13
Use JSON schema to define metadata.

zhenghuiwang:
#4 
This PR sets up basic structure and tests for future predefined metadata.

More context is in the README file in this PR and this https://github.com/kubeflow/metadata/issues/4#issuecomment-481064508

The intention is not to define any specific metadata schema, but show how everyone can contribute schemas and iterate on them in future PRs. (some examples are provided in `data_set.json`, `executor.json` and `execution.json`).

Please comment on it about how you think about this approach.

/cc @neuromage @jlewi @mvartakAtVerta @aronchick @terrytangyuan @ywskycn @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/10)


johnugeorge:
This will be very helpful. 
LGTM

schema/README.md
```diff
@@ -0,0 +1,30 @@
+# Metadata Schema
+
+This directory contains all predefined metadata types in the form of [JSON schema](https://json-schema.org). We expect every piece of metadata to have the following fields:
+
+- `id`of type _string_. Unique identifier assigned by this metadata service.
```
ywskycn: Do we want to use UUID here?

schema/alpha/artifacts/data_set.json
```diff
@@ -0,0 +1,55 @@
...
+    "required": [
+        "id",
+        "kind",
+        "namespace",
+        "apiversion",
+        "name",
+        "category",
+        "uri"
+    ],
+    "examples": [{
```
ywskycn: this example can be removed, right? as examples are put in separate directory.

jlewi:
/lgtm
/hold so @neuromage or someone else closely involved in metadata can approve
Is there a way to render the scheme prettily?

For example we use this site to render swagger.yaml
https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/kubeflow/kubeflow/0f8a3babb6e0f8425dd5dff9432a68889911366a/bootstrap/api/swagger.yaml

schema/README.md
zhenghuiwang: That's a good question.

UUID implies this field is 128-bit int. I want to let the implementation decide what kind of ID is  used. (e.g. it can be an incremental integer in mysql)

schema/alpha/artifacts/data_set.json
zhenghuiwang: It is duplicated from the example directory, but I want to keep it here also. So that the schema can be viewed with an example side by side.

zhenghuiwang:
> Couple of things -
> 
> 1. seems external efforts around metadata are following yaml syntax, whereas this one is using json - have we agreed here on json based schema? Or its mostly driven by MLMD implemenation currently?

It is not driven by MLMD, which has its own type system based on protobuf. It is very easy to convert between yaml syntax and json syntax. We can support both to define JSON schema.

> 2. an 'execution' is essentially a 'run' - any reason not to use the widely accepted nomenclature?

In the context of Kubeflow pipeline, 'run' is often used to refer a pipeline execution. Try to avoid overload 'run' with a pipeline and a single executable. Maybe we can use 'run' for both.

> 3. 'entity' and 'artifact' look very similar - what`s the rationale behind two here?
The difference is the artifact usually refers to a derived data somewhere( it has a `uri` to the data) while entity is some metadata in the system, which doesn't necessarily related to other data.

schema/alpha/artifacts/metrics.json
```diff
@@ -0,0 +1,6 @@
+
+{
```
mvartakAtVerta: Are metrics also a kind of artifact?

schema/alpha/containers/workspace.json
```diff
@@ -0,0 +1,5 @@
+{
```
mvartakAtVerta: The containers / workspace concept seems rather vaguely defined -- what is its purpose and does it belong in metadata?

schema/alpha/execution.json
```diff
@@ -0,0 +1,43 @@
...
+        {
+            "properties": {
+                "category": {
+                    "constant": "execution"
+                },
+                "executable_id": {
+                    "type": "string",
+                    "description": "the id of the executable"
+                },
+                "configuration": {
```
mvartakAtVerta: Is this supposed to be a catch all? It's great for flexibility but harder to operate on. Also brings up the question as to whether other entities should have a configuration property

aronchick:
I believe @vicaire had some opinions on YAML vs. JSON (his preference for YAML was why we felt more comfortable with YAML in [MLSpec](https://github.com/mlspec/MLSpec)) - if they are ABSOLUTELY interchangable, I'd prefer YAML as being more human readable. HOWEVER, if it slows our development at all, we should pick which ever one is faster!

vicaire:
Since Kubernetes is using yaml, and since Kubeflow is about making it easy to do ML on K8, I think using yaml would be a consistent choice.

animeshsingh:
> Since Kubernetes is using yaml, and since Kubeflow is about making it easy to do ML on K8, I think using yaml would be a consistent choice.

+1. Given we are mostly dealing with yaml in kubernetes world, and kuebflow is targeted for kube, it makes sense to stick to yaml based syntax. cc @mpvartak @vicaire @zhenghuiwang

zhenghuiwang:
@aronchick @vicaire @animeshsingh using yaml makes sense. I will change the format to yaml in this PR.

neuromage:
JSON seems fine to me.

The fact that we run on k8s shouldn't need to be a factor here, since we're building an application, that operates on ML metadata, not k8s metadata. JSON is a common enough interchange format, and also has the advantage that it can be easily and natively interpreted in any UI frontend that we build. If users have a YAML spec, then converting that to JSON should be easy anyway. As such, I think we should keep using JSON.

schema/alpha/artifacts/metrics.json
zhenghuiwang: Strictly speaking metrics is not artifact(no URI), it is easier to handle it as artifact.

zhenghuiwang:
To @mvartakAtVerta's comment
> * How will the JSON schema interface with the api in api.proto? That clearly seems to be using protos for definition. So are these schemas going to be translated into protos anyway? (would defeat the purpose of JSON schema then)

The schema is not going to translate to proto. Think there is a REST endpoint defined in api.proto: `CreateType(Request)` where `Request` is of type `message { string json_schema_str = 1;}` We deserialize the string to schema and use it to valid request to create metadata.

> * It's not clear what the purpose of containers is or how they should be used. If they are application specific, do they belong in metadata?
They are grouping concepts like workspace/project and experiment. They application specific metadata and we want to host them. E.g. Katib experiment spec.

> * Configuration (within execution) seems like a catch-all, 2 issues: (a) it is harder to operate on because it is so variable, is that expected; (b) if execution schema has it, when why not make it available on all entities? Would be worth looking into how k8s handles these arbitrary objects
I updated the pr to handle non-json data by [this guide](http://json-schema.org/understanding-json-schema/reference/non_json_data.html) in JSON schema

schema/alpha/containers/workspace.json
zhenghuiwang: replied inline

schema/alpha/execution.json
zhenghuiwang: replied inline

zhenghuiwang:
On YAML vs JSON as the schema format:

In term of how easy to be manipulated by programs, JSON is a clear winner, especially for JS application.

YAML is easy for human to read & write. For readability of schemas, it is not only about JSON syntax, but also the schema grammar. It's important to have readable docs for people who don't know JSON schema grammar, as @jlewi pointed out. So I added a script to generate markdown docs. ([sample doc](https://github.com/kubeflow/metadata/blob/2c02487126a070cb34c2e29065a2f452b0637741/schema/alpha/docs/entity.md)) Let me know if there is a better tool to autogen docs.

Unlike k8s config file, these schema files are not frequently changed manually. So I don't think it worth supporting YAML merely for making human editing easier. I would recommend using tool like Swagger editor to edit YAML schema and export it in JSON.

jlewi:
Folks have had plenty of time to chime in and no one has requested a hold. So I'm removing my hold and letting this go in. If anyone would like to discuss further. I suggest opening up an issue.

/lgtm
/approve
/hold cancel

neuromage:
/lgtm
/approve

Let's submit and we can iterate from there.

PR #15, opened 2019-04-23
Add JSON schema parser to the server & convert schemas to MLMD types

### LINKED ISSUE #20: Load predefined metadata into servers
part of #8 

Need to load JSON metadata spec into server during start up.

/assign @zhenghuiwang

zhenghuiwang:
Fixes #20

This PR adds a parser of JSON schema into the server:

- It parses json schema files into Go structs for easy manipulation: we can extract `id`, `type` `kind`, and `apiversion` from the schema etc.

- It compiles a set of schemas for validating JSON data.

- It collects a list of simple properties in a schema and register these properties as ArtifactType of MLMD.

Fixed Bazel build files to keep them in sync with `go.mod`.

/cc @neuromage @jlewi @mvartakAtVerta @terrytangyuan @ywskycn @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/15)


zhenghuiwang:
Test failure should be gone after the correct unit test config in https://github.com/kubeflow/metadata/pull/14 is landed

schemaparser/schemaset_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Fatalf("Unexpect simple properties. expected = %+v, got = %+v", expectedMap, sp)
+	}
+}
+
+func TestSimplePropertiesInExtendedSchema(t *testing.T) {
+	ss, err := NewSchemaSetFromFiles([]string{"testdata/base.json", "testdata/ext.json"})
+	if err != nil {
+		t.Fatalf("failed to parse schemas from files: %s", err)
+	}
+	id := "http://github.com/kubeflow/metadata/schemaparser/testdata/ext.json"
```
terrytangyuan: Consider moving this to a const

schemaparser/schemaset_test.go
```diff
@@ -0,0 +1,70 @@
+package schemaparser
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+var (
+	schemaDir = "../schema/alpha"
```
terrytangyuan: Change this to const instead of var?

schemaparser/schemaset_test.go
zhenghuiwang: Done. Thanks for the comments!

schemaparser/schemaset_test.go
zhenghuiwang: Done.

neuromage:
@zhenghuiwang Can we not add this to the server just yet? I'd rather we first settle on the right data model internally. Given that the server will not store using JSON schema, let's not expose this as first class concept just yet. I expect users to use the planned Python client to register types, like so:

```py
metadata_store.register_artifact_type(
    type_name='Model',
    schema={
       accuracy: types.Double,
       version: types.Int,
       ...
   })
```

This is vastly simpler than using a JSON schema.  

I'm ok with submitting this code as a helper tool in a separate directory, as a reference for later when we do want to support helping users convert an existing JSON schema to our data model.

What do you think?

/cc @hongye-sun

jlewi:
@neuromage What's the downside of adding this to the server? Isn't this an Alpha version of the server?

>  I expect users to use the planned Python client to register types, like so:

Why would types only be registered from python? Many Kubeflow applications (e.g. TFJob, Pipelines?) are written in go. So if those applications want to register types wouldn't they do it via RPC and not using the python client library.

neuromage:
@jlewi If they're written in Go, then they can just use the API protos directly. Or we can provide helper functions. My point is, the underlying data model is not going to be JSON schema, so exposing this from the get-go is overly complex, and isn't required in the initial version. And it introduces an impedance mismatch between the underlying storage model, and how users see their data schemas. I don't want to commit to the broader scope of supporting JSON schemas just yet.

zhenghuiwang:
@neuromage this library is not about the underlying data model, but adding a way for users to describe & validate JSON payload. Whatever schema you choose to describe the JSON payload, the promise of the REST server is the same: users can save a JSON object, have it validated, and later retrieve the same JSON object.

For the example you provide, it is also easy to write it in JSON schema(with more characters). But with a more complex object, you will probably need to extend your customized protobuf type system. If you are okay with making it as a separate tool, why it hurts to provide it directly to users, since JSON schema is a widely used way to describe JSON payload.

neuromage:
> ...users can save a JSON object, have it validated, and later retrieve the same JSON object.
No, this is a way for users to save a JSON schema that defines the type of the artifact. This is what I want to avoid for now. There will be a disconnect, since the types are not json schema internally.

jlewi:
@neuromage 

> And it introduces an impedance mismatch between the underlying storage model, and how users see their data schemas

Can't we auto-convert proto to json schemas? Are you saying that the underlying data can't be represented as json? 

> then they can just use the API protos directly. 

Are you expecting everyone to use protos and gRPC?

zhenghuiwang:
Updated the server startup logic to parse the predefined schemas and register them to MLMD. PTAL

@neuromage

schemaparser/schemajson_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	testSchemaFile = "testdata/ext.json"
+)
+
+func TestLoadSchemaJSON(t *testing.T) {
+	sj, err := schemaJSON(testSchemaFile)
+	assert.Nil(t, err, "Failed to load schema into SchemaJSON struct.")
```
neuromage: Consider not using testify, which is a really problematic library IMO.  Error messages are not helpful, and the test suite proceeds even after one asserts something should be non-nil, but is. I've seen this in pipelines a lot, which is I why I'd prefer not to use this library at all. The Go team recommends writing table-driven tests instead:

https://github.com/golang/go/wiki/TableDrivenTests

service/service.go
```diff
@@ -19,22 +19,106 @@ package service
...
+	typeNameToMLMDID map[string]int64
+
+	// TODO: Once MLMD type system supports schema annotations, the following two fields can be obtained by quering MLMD types.
+	typeNameToID map[string]string
+	schemaset    *schemaparser.SchemaSet
 }
 
 // NewService returns a new MetadataService.
-func NewService(mlmdClient mlmd.MetadataStoreServiceClient) *Service {
+func NewService(mlmdClient mlmd.MetadataStoreServiceClient, schemaRootDir string) (*Service, error) {
```
neuromage: Service should not need to know anything about schema root dir. It should be a standalone service, whose only concern is to export endpoints for serving Artifacts, Executions and Events. We could just call a separate function to do pre-registration.

Also, pre-registration should happen the same way users would call the endpoint. A standalone function takes your JSON schema, converts it to ArtifactType, and just calls the endpoint to register the latter instead. That way, Service is not aware of JSON schema or its parser.

service/service_test.go
```diff
@@ -18,10 +18,44 @@ import (
 	"context"
 	"testing"
 
+	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
 	pb "github.com/kubeflow/metadata/api"
+	mlmd "github.com/kubeflow/metadata/ml_metadata"
```
neuromage: With this package name, is Bazel build/test broken?

service/service.go
zhenghuiwang: I agree it is better to handle predefined & customized type the same way by calling an endpoint.

But now the problem is that we expose a different [ArtifactType](https://github.com/kubeflow/metadata/blob/master/api/api.proto#L43) from what current MLMD offers. So there needs another conversion between these two if we register types via endpoints. I'm trying to limit the work of this PR by handling predefined types first. It is an implementation details and we can move endpoint approach once we support customized types.

What do you think?

schemaparser/schemajson_test.go
zhenghuiwang: I have no preference and can use testing tables. Using it because it is shorter than writing `if err !=nil { t.Errorf(...)}` (You're right. here should be using t.Fatal)

service/service_test.go
zhenghuiwang: No, they work fine. The presubmit test checks Bazel build and test.

service/service.go
neuromage: I'm ok with registering types first before the other endpoint is ready. But I'm still not sure about this approach. I still see `category` and `artifactCategory`, which doesn't fit with the intended `ArtifactType` data model. Will you be removing all this later?

service/service.go
```diff
@@ -19,22 +19,106 @@ package service
...
 
 // Service implements the gRPC service MetadataService defined in the metadata
 // API spec.
 type Service struct {
-	mlmdClient mlmd.MetadataStoreServiceClient
+	mlmdClient       mlmd.MetadataStoreServiceClient
+	typeNameToMLMDID map[string]int64
+
+	// TODO: Once MLMD type system supports schema annotations, the following two fields can be obtained by quering MLMD types.
+	typeNameToID map[string]string
```
neuromage: Why do we need this id btw?

service/service.go
```diff
@@ -19,22 +19,106 @@ package service
...
+	schemaset    *schemaparser.SchemaSet
```
neuromage: I don't think Service needs this. All you need is a separate function, which could be in main like so:

```
func registerPredefinedSchema(ss *schemaParser, schemaRootDir string, mlmdClient *mlmd.MetadatStoreClient) {
...
}
```

This way, Service doesn't need to know about the JSON types. Later, when you move these types to ArtifactType definitions, there's less code to remove in Service. What do you think?

schemaparser/schemajson_test.go
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+package schemaparser
+
+import (
+	"testing"
+)
+
+var (
+	testSchemaFile = "testdata/ext.json"
```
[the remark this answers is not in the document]
zhenghuiwang: Done.

schemaparser/schemaset.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+		}
+		if parent.Ref != nil {
+			sj, err := ss.PropertyType((string)(*parent.Ref), name)
+			if err == nil {
+				return sj, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("faield to find property %q in %q", name, id)
```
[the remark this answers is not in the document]
zhenghuiwang: Done.

service/service.go
zhenghuiwang: Added a comment. this map the incoming type name to the id of corresponding schema and we can verify the incoming data with the schema.

service/service.go
zhenghuiwang: For current MLMD, there is no place to store the actual schema for fields. this schema set is the in memory storage for the schemas. so that we can use them to validate fields.

It is can be removed once MLMD can support schema annotations. but now it is needed. That's why I put a TODO above.

service/service.go
zhenghuiwang: The `category` is to indicate the predefined type is `artifactType` or `executionType`. It just makes the JSON schema self-explanatory. They won't be the properties. I added line 101.

service/service.go
```diff
@@ -19,22 +19,106 @@ package service
...
 	return &Service{
-		mlmdClient: mlmdClient,
+		mlmdClient:       mlmdClient,
+		typeNameToMLMDID: typeNameToMLMDID,
+		typeNameToID:     typeNameToID,
+		schemaset:        ss,
+	}, nil
+}
+
+func registerArtifactType(mlmdClient mlmd.MetadataStoreServiceClient, ss *schemaparser.SchemaSet, id string, name string) (*int64, error) {
```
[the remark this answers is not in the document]
zhenghuiwang: Done.

service/service_test.go
```diff
@@ -18,10 +18,44 @@ import (
...
+			return &mlmd.PutArtifactTypeResponse{
+				TypeId: &id,
+			}, nil
+		}).AnyTimes()
+	s, err := NewService(mockMLMDClient, schemaDir)
+	if err != nil {
+		t.Fatalf("Failed to create a new service: %v", err)
+	}
+	mustInclude := []string{
+		"namespaces/kubeflow.org/kinds/data_set/versions/alpha",
```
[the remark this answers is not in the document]
zhenghuiwang: I think we agree it will be useful to have several more specific types for data set. but it is still good to have a general type to capture any data if no suitable specific one can be used.

service/service.go
neuromage: SGTM, thanks that makes sense.

Issue #11, opened 2019-04-15
Installation of metadata service


zhenghuiwang:
As the first step of https://github.com/kubeflow/metadata/issues/9. We need A kustomize  component to install metadata service in KF cluster.

Make sure the metadata service is reachable. (We can ping the health endpoint of metadata service if KF is installed from Master). 

/cc @neuromage @jlewi @mvartakAtVerta @terrytangyuan @ywskycn

jlewi:
Should we split this up into sub issues?

zhenghuiwang:
This is a separate issue on the actual implementation #8. This one is about the service (a hello world implementation https://github.com/kubeflow/metadata/pull/3) can be installed from KF Master.

jlewi:
@zhenghuiwang How's this coming? Are we on track to complete this by end of week?

zhenghuiwang:
I didn't get a chance to work on it last week. But i will start it today or tomorrow. Should be able to finish it on time.

jlewi:
@zhenghuiwang Any update on this?

zhenghuiwang:
PRs are under reivew. The issue can be closed once the PRs are merged.

PR #16, opened 2019-04-25
Add Dockerfile for building service image

zhenghuiwang:
First step of #11 installing metadata as KF component.

Tested it locally.

/cc @neuromage @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/16)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# Build docker image at gcr.io/images-public/metadata:<tag>
+
 FROM golang:1.12
 
 ENV GO111MODULE on
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM golang:1.12
+
+ENV GO111MODULE on
+
+RUN apt-get update && apt-get -y install unzip patch && apt-get clean
+
+# Install Bazel
+RUN cd /tmp && \
+    wget -O /tmp/bazel-installer.sh https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-installer-linux-x86_64.sh && \
+    chmod +x bazel-installer.sh && \
+    ./bazel-installer.sh --user
+
+ENV PATH=/root/bin:${PATH}
+
+WORKDIR /go/src/github.com/kubeflow/metadata
+
+COPY . .
+
+RUN /bin/bash api/generate_proto.sh
+
+RUN cd server && go build ./...
+
+CMD ["./server/server"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,16 +4,6 @@ FROM golang:1.12
 
 ENV GO111MODULE on
 
-RUN apt-get update && apt-get -y install unzip patch && apt-get clean
-
-# Install Bazel
-RUN cd /tmp && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    chmod +x bazel-installer.sh && \
+    ./bazel-installer.sh --user
+
+ENV PATH=/root/bin:${PATH}
+
+WORKDIR /go/src/github.com/kubeflow/metadata
+
+COPY . .
+
+RUN /bin/bash api/generate_proto.sh
```
neuromage: This isn't needed, since we assume devs run this script whenever we update the API. For now, it may be sufficient to just build with `go build`.

zhenghuiwang: Generated code always concerns me. If we run this script before our presubmit tests, it slows down presubmit test every time. If we don't run this script for tests nor build, the proto and go code may be out of sync by mistake without notice.

What do you think?

zhenghuiwang:
> @zhenghuiwang can you also please help to get internal OSPO approval for releasing images? Thanks.

Thanks for point it out. How does OSPO work?

@jlewi Is there any example of other KF components that I can follow?

Dockerfile
zhenghuiwang: I think the presubmit test should check it, if  we want to make the gen code is in sync with the proto.

I removed this build step.

jlewi:
This isn't actually publishing an image its just the dockerfile so we're fine.

/lgtm
/approve

jlewi:
BTW you might eventually want to do a multi-stage build so the final image isn't polluted with the files used for building.  You might also want to add a GCB rule to build the image.

That can happen later though 

Here's an example Multi stage dockerfile
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/Dockerfile

And building it with GCB
https://github.com/kubeflow/kubeflow/blob/3e22dc1d0a8a32203e47831d97b26e56deabe41c/bootstrap/Makefile#L124

The GCB rule is nice because it avoids pulling the image to your local machine which can be slow

Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Build docker image at gcr.io/images-public/metadata:<tag>
+
+FROM golang:1.12
+
+ENV GO111MODULE on
+
+RUN apt-get update && apt-get -y install unzip patch && apt-get clean
+
+# Install Bazel
```
neuromage: No need for Bazel here.

zhenghuiwang: Removed.

zhenghuiwang:
> BTW you might eventually want to do a multi-stage build so the final image isn't polluted with the files used for building. You might also want to add a GCB rule to build the image.
> 
> That can happen later though
> 
> Here's an example Multi stage dockerfile
> https://github.com/kubeflow/kubeflow/blob/master/bootstrap/Dockerfile
> 
> And building it with GCB
> https://github.com/kubeflow/kubeflow/blob/3e22dc1d0a8a32203e47831d97b26e56deabe41c/bootstrap/Makefile#L124
> 
> The GCB rule is nice because it avoids pulling the image to your local machine which can be slow

Thanks for the suggestions. I will make these changes in future PRs.

PR #23, opened 2019-05-03
Add common gitignore patterns

terrytangyuan:
These are mostly adopted from https://github.com/github/gitignore and almost identical to what we have in kubeflow/common [here](https://github.com/kubeflow/common/blob/master/.gitignore), except for bazel related artifacts that this project use.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/23)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,71 @@
 # Bazel output artifacts
 bazel-*
 
-# VSCode settings.
-.vscode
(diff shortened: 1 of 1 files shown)


terrytangyuan:
Thanks @zhenghuiwang. Tests passed now.

PR #24, opened 2019-05-07
Add initial API and data model definitions.

neuromage:
This PR introduces some of the core data model concepts that will be
used by the metadata API server. All the definitions provided will be
subject to change as we evolve the API definitions in this early stage
of the project.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/24)


zhenghuiwang:
/lgtm

LGTM as the init setup. We can evolve this API over time.

neuromage:
> Looks good as the initial setup. Though are we missing "Container" that's mentioned [here](https://github.com/kubeflow/metadata/tree/master/schema#metadata-schema)?

@terrytangyuan Thanks for bring that up. I don't think we'll be exposing this concept just yet. We'll most likely focus on initial groupings of artifacts and executions through pre-defined concepts like `Workspaces`.

terrytangyuan:
> > Looks good as the initial setup. Though are we missing "Container" that's mentioned [here](https://github.com/kubeflow/metadata/tree/master/schema#metadata-schema)?
> 
> @terrytangyuan Thanks for bring that up. I don't think we'll be exposing this concept just yet. We'll most likely focus on initial groupings of artifacts and executions through pre-defined concepts like `Workspaces`.

Sure. Sounds good.

PR #25, opened 2019-05-07
Add ML MetaData as dependency

### LINKED ISSUE #18: Add MLMD server
part of #8 

Introduce MLMD as dependency. Try to avoid using it as a library. Otherwise the standard Go tool chain will break, since it is written in C++.

/assign @zhenghuiwang

zhenghuiwang:
Fixes #18 

This PR introduce MLMD GRPC client as dependency. We don't use it as a library since it is written in C++ and using it as library breaks the standard Go tools.

- Add Makefile to update the MLMD client proto and generate new client stub.
- Add Dockerfile for building an image of MLMD server.
-  Init  MLMD client during server start.

Tested it locally: verified the client can talk to the MLMD server.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/25)


neuromage:
@zhenghuiwang I think we need to use this as a library. We'll need to hack in join tables at the start, and ensure updates happen in the same transaction. Also, launching two services instead of one will add complexity to the deployment and testing part, especially in the beginning. Agreed that it's inconvenient in that it breaks standard Go tooling, but I think we can look into fixing that separately or later. It shouldn't be a reason for using the grpc service since that's a development concern.

zhenghuiwang:
>  We'll need to hack in join tables at the start, and ensure updates happen in the same transaction. 

I don't see why it is a problem of RPC calls but not library calls. If you want to join tables in a transaction using mySQL, then our server will be issuing mySQL queries directly to MySQL servers for tables MLMD manages + other tables. Another way is for our server to manage locks for MLMD calls + MySQL queries, but how to call MLMD is irrelevant. 

> Also, launching two services instead of one will add complexity to the deployment and testing part, especially in the beginning.

The GPRC client is an interface, which can be mocked for unit tests easily.

Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+test:
+	go test ./...
+
+hello:
+	curl localhost:8080/api/v1/resource/my-resource
+
+mlmd-proto:
+	rm -rf /tmp/ml-metadata && \
+	git clone https://github.com/google/ml-metadata.git /tmp/ml-metadata && \
```
terrytangyuan: Should we tag a particular version of MLMD? It would be easier when upgrading.

neuromage:
@zhenghuiwang Sorry, but my initial point still stands. The reason for having two services here seems to be to aid the developer, rather than for creating two independent micro-service. If I figure out a way to not break Go standard tooling, would you be ok with not using the grpc service?

jlewi:
I think there is a lot of value in using the go tool chain as opposed to using bazel.
bazel can be a big barrier to new developers.

Why is deploying two separate services problematic? Can't you just deploy it in a side car?

animeshsingh:
I would vote against bazel as well - as I mentioned before in one of the issues, its not used much outside of google anyways

animeshsingh:
Another question - should this be a dependency or a plugin? At best the default option? And we can choose ModelDB if desired?

Makefile
zhenghuiwang: Add a tag env.

zhenghuiwang:
> Another question - should this be a dependency or a plugin? At best the default option? And we can choose ModelDB if desired?

It is a dependency for the implementation with MLMD. The metadata service API will be a generated GRPC Go interface. Different implementations for that interface can be plugged in.

zhenghuiwang:
@neuromage I think the side car approach that Jeremy mentioned addresses the concern well about exposing the underlying MLMD servers. (I found [this article](https://www.mirantis.com/blog/multi-container-pods-and-container-communication-in-kubernetes/) has good explanations and a very relevant example)

Makefile
```diff
@@ -0,0 +1,26 @@
...
+hello:
+	curl localhost:8080/api/v1/resource/my-resource
+
+mlmd-proto:
+	rm -rf /tmp/ml-metadata && \
+	git clone --branch $(TAG) https://github.com/google/ml-metadata.git /tmp/ml-metadata && \
+	rm -rf ml_metadata/proto && \
+	mkdir ml_metadata/proto && \
+	cp /tmp/ml-metadata/ml_metadata/proto/*.proto ml_metadata/proto && \
+	protoc -I . ml_metadata/proto/*.proto --go_out=plugins=grpc:. && \
```
neuromage: Let's not use developer-installed protoc and building from ml_metadata head. Instead, let's use Bazel for proto generation instead. This ensures the build is hermetic and we won't see proto files oscillating due to dev setup (this was a problem in pipelines). See the example I did using the script + BUILD rule for how to do this. Alternately, if you're not sure how to do it, leave a TODO here for me and I'll change this part later.

service/service.go
```diff
@@ -21,11 +21,21 @@ import (
 	"errors"
 
 	pb "github.com/kubeflow/metadata/api"
+	mlmd "github.com/kubeflow/metadata/ml_metadata"
```
neuromage: This is weird to me. These protos are third party, and we're building and using them under our package name, which seems wrong. How do we keep the ml_metadata package name (and not use ours here)?

server/main.go
```diff
@@ -42,10 +44,16 @@ func main() {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	rpcEndpoint := fmt.Sprintf("%s:%d", *host, *rpcPort)
+	conn, err := grpc.Dial(*mlmdAddr, grpc.WithInsecure())
+	if err != nil {
+		glog.Fatalf("fail to dial to MLMD server: %v", err)
```
neuromage: Nit: "Failed to connect to MLMD server:...."

server/main.go
```diff
@@ -34,6 +35,7 @@ var (
 	host     = flag.String("host", "localhost", "Hostname to listen on.")
 	rpcPort  = flag.Int("rpc_port", 9090, "RPC serving port.")
 	httpPort = flag.Int("http_port", 8080, "HTTP serving port.")
+	mlmdAddr = flag.String("mlmd_address", "127.0.0.1:9090", "the server address of ml_metadata in the format of host:port")
```
neuromage: Nit: Capitalize "The server address..."

service/service.go
zhenghuiwang: The goal of the package is to compile a Go client for MLMD. Maybe 
 changes the package path to github.com/kubeflow/metadata/`clients/ml_metadata` makes it clear.

WDYT?

Makefile
zhenghuiwang: SG. I added a TODO

server/main.go
zhenghuiwang: Done

server/main.go
zhenghuiwang: Done.

service/service.go
neuromage: I guess I'm asking if we can use package name like `github.com/ml-metadata/ml_metadata/ml_metadata_store_proto` or something along those lines. Is this possible or not really when using third party protos in Go? 

If this is not possible from the client side (and I suspect it's not), then I think the right solution here could be to get ml-metadata to export the Go proto files. So that way, we can use consistent package names that works across both Bazel and Go. 

I will lgtm this to unblock you, but I think we can look into publishing those Go proto files upstream and rely on them instead once it's ready. Does this work for you?

zhenghuiwang: > If this is not possible from the client side (and I suspect it's not), then I think the right solution here could be to get ml-metadata to export the Go proto files.

I think that is the only way to define `github.com/ml-metadata/ml_metadata/ml_metadata_store_proto`: move this compiled Go package here to MLMD's github repo. If this happens, we can import it as other Go packages, which is cleaner solution.

I will land this PR for now. Thanks!

PR #26, opened 2019-05-09
Init type spec for model, metrics and workspace

### LINKED ISSUE #17: Create Metadata spec for model, metrics, workspace and etc
part of #4 

Create the basic type spec for model, metrics, execution, workspace etc so that we can save/retrieve these data directly.

It is not intended to be comprehensive for all use cases.

/assign @zhenghuiwang

zhenghuiwang:
Fixes #17 

This PR adds the initial type specification for model, metrics and workspace. These initial specs should be sufficient for our v0.6 CUJs.

For easier review, generated schema docs can be found [here](https://github.com/zhenghuiwang/metadata/tree/kfmd-predefined-types/schema/alpha/docs).

Changes:
- Adds type spec for model, metrics and workspace.
- Moves all examples into the schema `examples` field so that the generated docs can show the examples.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/26)


schema/alpha/docs/containers/workspace.md
```diff
@@ -5,7 +5,281 @@
...
+```
+
+#  Properties
+
+| Property | Type | Required | Nullable | Defined by |
+|----------|------|----------|----------|------------|
+| [annotations](#annotations) | `object` | Optional  | No | [entity](../entity.md#annotations) |
+| [apiversion](#apiversion) | `string` | **Required**  | No | [entity](../entity.md#apiversion) |
+| [category](#category) | `enum` | **Required**  | No | [entity](../entity.md#category) |
+| [create_time](#create_time) | `string` | Optional  | No | [entity](../entity.md#create_time) |
```
terrytangyuan: Perhaps consider renaming to either "created_at" or "creation_time"?

zhenghuiwang: All these names make sense, but we have choose one convention. I'm following the Google could API design guide line [here].(https://cloud.google.com/apis/design/naming_convention#time_and_duration)

terrytangyuan: Okay sounds good as long as we follow a particular convention.

schema/alpha/artifacts/artifact.json
```diff
@@ -1,6 +1,23 @@
 {
     "$id": "http://github.com/kubeflow/metadata/schema/alpha/artifacts/artifact.json",
     "$schema": "http://json-schema.org/draft-07/schema#",
+    "examples": [{
+        "category": "artifact",
+        "id": "123",
+        "kind": "model",
```
animeshsingh: Is this a free flowing list of key-value pairs? Not checked against a specific schema e.g. I can define my own "kind"?

schema/alpha/artifacts/executable.json
```diff
@@ -1,6 +1,31 @@
...
+            }
+        ],
+        "kind": "my-traning-code",
+        "namespace": "my-company.org",
+        "output_type": [
+            {
+                "apiversion": "v1",
+                "category": "artifact",
+                "kind": "model",
+                "namespace": "my-company.org"
```
animeshsingh: I would assume namespace is optional

schema/alpha/artifacts/executable.json
```diff
@@ -1,6 +1,31 @@
...
+            }
+        ],
+        "apiversion": "v1",
+        "uri": "file://local_file.py",
```
animeshsingh: In most real world scenarios, its not a straightforward uri for file location

schema/alpha/artifacts/model.json
```diff
@@ -1,7 +1,86 @@
-
 {
-
     "$id": "http://github.com/kubeflow/metadata/schema/alpha/artifacts/model.json",
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "description": "TODO: Define the metadata about a model."
+    "examples": [{
```
animeshsingh: Is this something an end-user is expected to craft and submit?

schema/alpha/artifacts/artifact.json
zhenghuiwang: > Is this a free flowing list of key-value pairs? 

It is right now. But you can constrain it to only known properties by setting`  "additionalProperties": false` in this schema.
http://json-schema.org/understanding-json-schema/reference/object.html#properties

> Not checked against a specific schema e.g. I can define my own "kind"?

The properties of this example is checked against the `allOf` section in this file. For any known properties in the `allOf`, it has to satisfy all constraints.

You can define you own `model`. e.g. see `schema/alpha/artifacts/model.json` file, which defines  `model` artifact for `kubeflow.org`. You can add a schema for your own model type.

schema/alpha/artifacts/executable.json
zhenghuiwang: Yes, `output_type` now is unconstrained as it is defined in the following section as 
```
"output_type": {
                    "items": {
                        "type": "object"
                    },
                    "type": "array"
                },
```

schema/alpha/artifacts/executable.json
zhenghuiwang: Yes, this can be any string as long as the systems/users know how to use it. [more examples](https://github.com/kubeflow/metadata/blob/master/schema/alpha/docs/artifacts/artifact.md#uri-examples)

schema/alpha/artifacts/model.json
zhenghuiwang: End users are not expected to submit these type definitions, but the platform engineers may need to define them and register them if they want to have types not listed here.

This PR adds these types for `kubeflow.org` so that most kubeflow users can directly save/retrieve data of these types. But if a company wants to customized their own `model` or `data_set` etc, they can define their own schemas.

jlewi:
@zhenghuiwang looks like @terrytangyuan LGTM'd this. Is there anything blocking this PR? You should be able to approve it since you are in the OWNERs file.

Issue #19, opened 2019-04-29
Add endpoints to Create and Read metadata


zhenghuiwang:
part of #5 

Add endpoints to CRUD metadata for predefined types.

/assigne @zhenghuiwang

jlewi:
@zhenghuiwang Any update on this?

jlewi:
@neuromage Could you provide an update please on this issue? What is the remaining work to be done?

jlewi:
@neuromage ping?

neuromage:
Fixed with #49

PR #30, opened 2019-05-21
Beginnings of ArtifactType API

neuromage:
This PR begins to implement methods to serve ArtifactType. 

This PR also switches to using a single server that uses MLMD as the backend (instead of the previous micro-service architecture). Bazel is the primary build tool due to the dependency on C++, but `go build` will work once Bazel has run once to produce output artifacts used during linking phase.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/30)


neuromage:
@zhenghuiwang  Tests are incomplete, and it's likely presubmit will fail. I will fix this tomorrow. For now, this should give you an idea of how I've split the server and how to use `go build` to build a single service.

third_party/ml_metadata/metadata_store/metadata_store_go_wrap/metadata_store_go_wrap.go
```diff
@@ -0,0 +1,580 @@
...
+extern uintptr_t _wrap_new_Status__SWIG_1_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1, uintptr_t arg2);
+extern uintptr_t _wrap_new_Status__SWIG_2_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1);
+extern _Bool _wrap_Status_ok_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1);
+extern uintptr_t _wrap_Status_code_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1);
+extern swig_type_38 _wrap_Status_error_message_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1);
+extern void _wrap_delete_Status_metadata_store_go_wrap_79669df856d5e5b8(uintptr_t arg1);
+#undef intgo
+
+#cgo CXXFLAGS: -std=c++11
+#cgo CXXFLAGS: -I${SRCDIR}/../../../../bazel-metadata/external/com_google_absl
```
neuromage: @zhenghuiwang These are the flags I added to enable `go build` to work with `cgo`.

service/service.go
```diff
@@ -20,116 +20,218 @@ import (
...
+}
+
+func toMLMDArtifactType(in *pb.ArtifactType) (*mlpb.ArtifactType, error) {
+	res := &mlpb.ArtifactType{
+		Id:         proto.Int64(in.Id),
+		Properties: make(map[string]mlpb.PropertyType),
+	}
+
+	for k, v := range in.TypeProperties {
+		if strings.HasPrefix(k, "kf_") {
```
zhenghuiwang: can we use `__kf_` as the prefix to avoid possible collisions?

neuromage: Done.

neuromage:
Part of addressing issue #19

Issue #40, opened 2019-05-29
Publish metadata python client to PyPI


zhenghuiwang:
part of #6

Need to support this for users to pip install in notebook environment.

PR #41, opened 2019-05-30
release python sdk

gaoning777:
It builds a python package that we can upload to the pypi.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/41)


gaoning777:
prepare for releasing the sdk: https://github.com/kubeflow/metadata/issues/40

sdk/python/setup.py
```diff
@@ -0,0 +1,49 @@
...
+
+setup(
+    name=NAME,
+    version=VERSION,
+    description='KubeFlow Metadata SDK',
+    author='google',
+    install_requires=REQUIRES,
+    packages=[
+        'kfmd',
+        'kfmd.swagger_client',
```
zhenghuiwang: Should we check in the swagger client? (if so, we can do it later)

If we don't check in the generated code in the source repo, should we generate it before publishing the package?

gaoning777: The current build.sh requires the swagger client to be generated before the kfmd package can be built. In this case, the generated swagger client is not checked in the source repo but the code will be in the kfmd package.

zhenghuiwang: Sounds good. I'm leaning towards checking in the generated code under source control but we can do it later.

gaoning777:
Would you mind also approve since I'm not in the owners list?

PR #42, opened 2019-05-30
Add MySql password flag

zhenghuiwang:
Tested it locally.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/42)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ TAG ?= master
 build:
 	bazel build -c opt --define=grpc_no_ares=true //...
 run:
-	bazel run --define=grpc_no_ares=true //server -- --logtostderr --mysql_service_port=3306 --mysql_service_user=guest --mlmd_db_name=metadb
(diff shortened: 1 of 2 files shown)


neuromage:
/lgtm

How is the password set initially? What does `%password%` get substituted with?

zhenghuiwang:
> How is the password set initially? What does `%password%` get substituted with?

In production, the mysql container and metadata server container will share the same [secret](https://kubernetes.io/docs/concepts/configuration/secret/) when starting both services.

neuromage:
> > How is the password set initially? What does `%password%` get substituted with?
> 
> In production, the mysql container and metadata server container will share the same [secret](https://kubernetes.io/docs/concepts/configuration/secret/) when starting both services.

Is there a connection between secrets and the `%password%` substitution? Sorry if it's obvious, but can you explain how this would work to me?

zhenghuiwang:
> Is there a connection between secrets and the `%password%` substitution? Sorry if it's obvious, but can you explain how this would work to me?

I should have been more specific: the k8s secret is populated to container as the env variable and you can set the cmd flag like this:
https://github.com/kubeflow/manifests/blob/master/metadata/base/metadata-deployment.yaml#L28

PR #50, opened 2019-06-12
Update python SDK to adapt new artifact protobuf

zhenghuiwang:
This PR updates the python SDK for recent artifact protobuf changes:

- Move fields `name`, `create_time` into artifact property map; move `workspace`, `run` into customized properties.
- Change swagger code generation to use [openapi generator](https://github.com/openapitools/openapi-generator), which fixes the issue that current code generator generates invalid python code (for `artifactStructType` proto) with circular references.
- Reorder some endpoints so that they won't shadow each other.

**Tests**
Passed tests `sdk/python/tests/test.py ` and `make_request.sh` locally

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/50)


sdk/python/kfmd/metadata.py
```diff
@@ -82,21 +83,26 @@ def log(self, artifact):
...
     """
 
     # TODO(zhenghui): log this artifact as the input or output of an execution.
     serialization = artifact.serialization()
-    if serialization.workspace.name == None:
-      serialization.workspace.name = self.workspace.name
+    if serialization.custom_properties == None:
+          serialization.custom_properties = {}
+    serialization.custom_properties[
+      "__workspace__"] = openapi_client.MlMetadataValue(
```
neuromage: should we use `__kf_workspace` ? And check that clients never use this property accidentally?

zhenghuiwang: Done.

zhenghuiwang:
I've done my edits. It is ready for your review @gaoning777

sdk/python/kfmd/metadata.py
```diff
@@ -82,21 +86,32 @@ def log(self, artifact):
     This method expects `artifact` to have
       - ARTIFACT_TYPE_NAME stirng field the form of
```
gaoning777: nit: stirng to string

sdk/python/kfmd/metadata.py
```diff
@@ -82,21 +86,32 @@ def log(self, artifact):
...
-      serialization.workspace.name = self.workspace.name
+    if serialization.custom_properties == None:
+          serialization.custom_properties = {}
+    if self.WORKSPACE_PROPERTY_NAME in serialization.custom_properties:
+          raise ValueError("custom_properties contains reserved key %s"
+                           % self.WORKSPACE_PROPERTY_NAME)
+    if self.RUN_PROPERTY_NAME in serialization.custom_properties:
+      raise ValueError("custom_properties contains reserved key %s"
+                       % self.RUN_PROPERTY_NAME)
+    serialization.custom_properties[
```
gaoning777: Would you mind increasing the linebreak length? This is a little hard to read.

neuromage: Actually, I prefer it this way. I recommend we follow the style guide prescription here:
http://google.github.io/styleguide/pyguide.html#32-line-length

sdk/python/kfmd/metadata.py
zhenghuiwang: good catch! I will correct it in subsequence pr to avoid needing a lgtm label again.

zhenghuiwang:
Thanks for the reviews!

/approve

PR #52, opened 2019-06-13
Add db options in the server

gaoning777:
This PR adds the sqlite option in the service. Currently, it only supports a fake sqlite in-memory instance. 

Tests:
Passed tests sdk/python/tests/test.py and make_request.sh locally
except that the ListArtifactTypes shows "cannot find type". The comment(https://github.com/kubeflow/metadata/blob/9f90726d94da6bef9ca9eb8c951623ad00135d85/service/service.go#L175) shows that the server does not support list artifact types.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/52)


server/main.go
```diff
@@ -49,16 +50,28 @@ var (
...
+					Host:     mySQLServiceHost,
+					Port:     proto.Uint32(uint32(*mySQLServicePort)),
+					Database: mlmdDBName,
+					User:     mySQLServiceUser,
+					Password: mySQLServicePassword,
+				},
+			},
+		}
+	} else {
+		glog.Fatalf("Unknown mlmd_db_type %s: please choose from [mysql, sqlite]", *mlmdDBType)
```
neuromage: %s => %q so you get it quoted

server/main.go
```diff
@@ -49,16 +50,28 @@ var (
...
-	cfg := &mlpb.ConnectionConfig{
-		Config: &mlpb.ConnectionConfig_Mysql{
-			&mlpb.MySQLDatabaseConfig{
-				Host:     mySQLServiceHost,
-				Port:     proto.Uint32(uint32(*mySQLServicePort)),
-				Database: mlmdDBName,
-				User:     mySQLServiceUser,
-				Password: mySQLServicePassword,
+	var cfg *mlpb.ConnectionConfig
+	if *mlmdDBType == "sqlite" {
```
neuromage: This is not sqlite, this is the in-memory fake option. Change sqlite to `in-memory` instead? You may want to also provide the `sqlite` option as well. In which case, I would change the if else to use Go `switch` statement instead.

server/main.go
gaoning777: SG.

server/main.go
gaoning777: Will do.

gaoning777: Done.

server/main.go
```diff
@@ -41,24 +41,49 @@ var (
 	httpPort      = flag.Int("http_port", 8080, "HTTP serving port.")
 	schemaRootDir = flag.String("schema_root_dir", "schema/alpha", "Root directory for the predefined schemas.")
 
+	mlmdDBType					 = flag.String("mlmd_db_type", "in-memory", "Database type to use when creating MLMD instance. Supported options: mysql, in-memory, sqlite")
```
zhenghuiwang: Can you make "mysql" as the default value? Since we may have some command/config, which is using mysql, not aware of this new flag?

gaoning777: Done.

server/main.go
```diff
@@ -41,24 +41,49 @@ var (
 	httpPort      = flag.Int("http_port", 8080, "HTTP serving port.")
 	schemaRootDir = flag.String("schema_root_dir", "schema/alpha", "Root directory for the predefined schemas.")
 
+	mlmdDBType					 = flag.String("mlmd_db_type", "mysql", "Database type to use when creating MLMD instance. Supported options: in-memory, mysql, sqlite")
 	mlmdDBName           = flag.String("mlmd_db_name", "mlmetadata", "Database name to use when creating MLMD instance.")
 	mySQLServiceHost     = flag.String("mysql_service_host", "localhost", "MySQL Service Hostname.")
 	mySQLServicePort     = flag.Uint("mysql_service_port", 3306, "MySQL Service Port.")
 	mySQLServiceUser     = flag.String("mysql_service_user", "root", "MySQL Service Username.")
 	mySQLServicePassword = flag.String("mysql_service_password", "", "MySQL Service Password.")
+	sqliteFilenameUri		 = flag.String("sqlite_filename_uri", "mlmetadata", "Sqlite Filename URI")
```
neuromage: Maybe run goformat on the file before submitting.

gaoning777: Done. Could you re-lgtm?

PR #53, opened 2019-06-14
Add list artifacts by type into Python SDK

zhenghuiwang:
This PR adds a method to list artifacts by type in a workspace, which can be later used to display new models trained in a Notebook.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/53)


zhenghuiwang:
Thanks for the review!

/approve

PR #62, opened 2019-06-24
Updates client library from new spec and creates Artifact and Model-specific Details Views

prodonjs:
- Updates client library generated code to the latest specification.
- Adds call to retrieve Artifact Types and merge with Artifacts on List page.
- Updates Router configuration to conform to API format (Artifact Type
  and ID) and creates specific Route for Model Details.

/cc @rileyjbauer

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/62)


frontend/src/pages/ArtifactList.tsx
```diff
@@ -66,28 +73,30 @@ class ArtifactList extends Page<{}, PipelineListState> {
...
-          type,
-          a.uri,
-          a.custom_properties![CustomProperties.WORKSPACE] ?
-            a.custom_properties![CustomProperties.WORKSPACE].string_value : '',
-          formatDateString(a.properties!.create_time!.string_value),
-        ],
-      };
-    });
+    const rows: Row[] = this.state.artifacts
+      .filter((a) => !!a.properties) // We can't show much without properties
```
rileyjbauer: Does this prevent the artifact from showing up at all in the list? That doesn't seem desirable without at least allowing users to remove the filter if they choose

frontend/src/pages/ArtifactList.tsx
```diff
@@ -125,12 +134,12 @@ class ArtifactList extends Page<{}, PipelineListState> {
...
-      // const response = await this.api.metadataService.listArtifactTypes();
-      // return new Map(
-      //   response.artifact_types!.map((at) => [at.id!, at])
-      // );
+      const response = await this.api.metadataService.listArtifactTypes();
+      return new Map(
+        response.artifact_types!.map((at) => [at.id!, at])
+      );
     } catch (err) {
+      // TODO: Show error since we won't be able to resolve links?
```
rileyjbauer: Yes, I think a banner error/warning would be good for this path. We can still show the artifacts themselves because maybe that will be useful, but there should be an indication to the user that something isn't right

frontend/src/pages/ModelDetails.test.tsx
```diff
@@ -5,62 +5,52 @@ import {shallow, ShallowWrapper, ReactWrapper} from 'enzyme';
...
-        custom_properties: {
-          __kf_workspace__: {string_value: 'workspace-1'},
-          __kf_run__: {string_value: '1'},
-        },
+  const mockGetArtifact = jest.spyOn(
+    Api.getInstance().metadataService, 'getArtifact');
+  const fakeArtifactResponse: ApiGetArtifactResponse = {
+    artifact:
+    {
+      id: "1",
```
rileyjbauer: nit: single quotes

frontend/src/pages/ArtifactList.tsx
prodonjs: It does prevent them from displaying. I put this in place on Friday because the service was returning some items that didn't seem to be populated completely. I'm not able to get a valid response from the service at this point, but I'll retry to see if this is still the case.

frontend/src/pages/ArtifactList.tsx
prodonjs: Added a banner message. We won't be able to show any links in that case so I'll work to get that change added as well.

frontend/src/pages/ModelDetails.test.tsx
prodonjs: Done.

prodonjs:
@rileyjbauer Made some small changes here. I think I am going to try to refactor this a bit into a generic artifact details page which render logic in place to handle models specifically. That would give us a bit more flexibility going forward and allow us to at least provide links for all artifact types. I'll probably do that in a separate PR though so as not to make this too big for review.

prodonjs:
@rileyjbauer I've made the necessary changes on this PR to be able to link to all artifact types, but show the specific [layout](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZUYQ_HerEEyO/files/MCEJu8Y2hyDScfbmC94M0URb19DWmvsbGa4) indicated for Models when appropriate. I also updated the Dockerfile with a few small changes to streamline the image size. I'd like to run the unit tests there, but right now there are issues with the date strings based on our use of `toLocaleString` in Utils.

I also tried to deploy to @zhenghuiwang 's cluster by adding the container to the existing [Deployment](https://pantheon.corp.google.com/kubernetes/deployment/us-central1-a/kubeflow-v05/kubeflow/metadata-deployment?project=zhenghui-kubeflow&organizationId=433637338589&tab=yaml&deployment_overview_active_revisions_tablesize=50&duration=P1D&pod_summary_list_tablesize=20&service_list_datatablesize=20), but that's probably not the right thing to do here as it's not allowing me to set up port-forwards and test it live.

In any case, here's a [screencast](https://screencast.googleplex.com/cast/NTQ1NTQyMjUzMjA5MTkwNHw4NmFjNmY2NS0xNw) of the current state.

prodonjs:
With the change you made yesterday on the API server, I was able to push that image into @zhenghuiwang 's cluster and also create a deployment that seems to work properly when talking to the backend.

`kubectl port-forward $(kubectl get pods -l 'kustomize.component=metadata,component=ui' -o name) 8000:3000` works correctly when I load the UI on http://localhost:8000.

frontend/src/components/ArtifactInfo.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  public render(): JSX.Element {
+    const {artifact} = this.props;
+    const createTime = formatDateString(
+      getArtifactProperty(artifact, ArtifactProperties.CREATE_TIME) || '');
+    return (
+      <section>
+        <h1 className={commonCss.header}>{this.props.typeName} info</h1>
+        <dl className={css.artifactInfo}>
+          <div className={css.field}>
```
rileyjbauer: Doesn't have to be done in this PR (or at all), but each of these looks similar enough to be generated by a function. The only things that are changing are the `term` and `value`

frontend/src/components/ModelInfo.tsx
```diff
@@ -14,27 +14,12 @@
...
-    color: color.grey,
-  },
-  value: {
-    color: color.secondaryText,
-  }
-});
+import {classes} from 'typestyle';
+import {commonCss} from '../Css';
+import {getArtifactProperty, logger} from '../lib/Utils';
+import {css} from './ArtifactInfo';
```
rileyjbauer: This should probably be put in a more common location instead of pulled from a different component

frontend/Dockerfile
```diff
@@ -3,40 +3,38 @@ FROM node:12.3.1 as build
...
+# Write commit and build date files and generate the dependency licenses files
+# (one for the UI and one for the webserver), concatenate them to one file
+# under ./src/server
 RUN mkdir -p ./server/dist && \
-    echo ${COMMIT_HASH} > ./server/dist/COMMIT_HASH && \
-    echo ${DATE} > ./server/dist/BUILD_DATE
-
-# Generate the dependency licenses files (one for the UI and one for the webserver),
-# concatenate them to one file under ./src/server
-RUN npm i -D license-checker
```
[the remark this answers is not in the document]
prodonjs: I just added to the `devDependencies` so it would get installed with the rest of the package.

frontend/src/components/ArtifactInfo.tsx
prodonjs: Agreed. This whole thing is pretty generic right now so it could easily be done in a map.

frontend/src/components/ModelInfo.tsx
prodonjs: Yeah I wasn't quite sure what the best pattern for doing that would be since the ModelInfo was just adding terms to its parent. Given your comment above, maybe the best thing to do is just make a functional component that for the wrapper div, dt, and dd elements. For the sake of time (and the fact that this has other changes coupled with it), I'd like to push that off to a subsequent change but I'll make a TODO for it if you'd like.

PR #65, opened 2019-06-26
Add endpoint for connecting artifacts to executions as input/output

zhenghuiwang:
This PR exports two endpoints to _create_ and _search_ MLMD's `event` - a input/output relation between artifacts and executions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/65)


zhenghuiwang:
It is ready for your review.
/assign @neuromage

api/service.proto
```diff
@@ -287,4 +298,23 @@ service MetadataService {
...
+  rpc CreateEvent(CreateEventRequest)
+      returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      post: "/api/v1alpha1/events"
+      body: "event"
+    };
+  }
+
+  // Search events based on an artifact or execution id.
+  rpc SearchEvents(SearchEventsRequest)
```
neuromage: Why not use a ListEvents call instead?

zhenghuiwang: https://cloud.google.com/apis/design/standard_methods#list

My understanding is `list` is for searching child resources in a collection, e.g. books in a shelf. `search` method is for searching related resources, which doesn't need to be a parent-child relation.

neuromage: Events is a resource, so you can list it. The question then is how would one filter this list. You can have an optional `filter` field with an expression that can specify how this should be restricted. So filter could be execution=? or artifact=? This is probably the more flexible option with richer possibilities for lineage search queries later (maybe, I haven't thought this through properly yet).

zhenghuiwang: I changed it to `list` method, as I don't think the difference matters that much.

`event` is a tuple `(artifact_id, execution_id,type)`. This method list all tuples contains certain artifact_id/execution_id.  The obvious filtering is by `type` (you don't want to filter by artifact_id/execution_id as that what are being searched for)

But I think more useful filtering logic is to further fetch the artifact/execution and filter the events based on the properties of artifact/execution. E.g. given an execution_id, find the events that contains artifact of type `dataset`.

I think it needs further thoughts and should be designed together with `ListArtifacts`/`ListExecutions`.

So how about addressing it in future changes?

zhenghuiwang:
Thanks for the review!

/approve

PR #80, opened 2019-07-01
Add homepage entry to allow serving from non-root folder

prodonjs:
This enables the Metadata UI to be served from the `/metadata` virtual service path within a deployed Kubeflow cluster.

gcr.io/kubeflow-images-public/metadata-frontend:v0.1.7 has been pushed with this change in it.

Sample cluster at:
https://kubeflow-v-n00.endpoints.zhenghui-kubeflow-2.cloud.goog/_/metadata/#/artifacts

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/80)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -57,6 +57,7 @@
       "last 1 safari version"
     ]
   },
+  "homepage": "./",
   "jest": {
     "collectCoverageFrom": [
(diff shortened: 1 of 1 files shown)


zhenghuiwang:
Thanks for the fix!

/lgtm
/approve

PR #82, opened 2019-07-03
Adds grouping/expansion to Artifacts List view

### LINKED ISSUE #71: [UI] Add grouping with expansion for Artifacts List view
/kind feature
/area front-end

Group artifacts by type and name showing the most recent artifact by creation time first.

[Mocks](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZUYQ_HerEEyO/files/MCEJu8Y2hyDScV4JnFyjZQDQkSM1yDiF2do)

prodonjs:
Adds grouping and expansion to Artifacts List view in accordance with [UX mocks](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZUYQ_HerEEyO/files/MCEJu8Y2hyDScV4JnFyjZQDQkSM1yDiF2do).

**Which issue(s) this PR fixes**
Fixes #71

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/82)

### DIFF
--- a/frontend/src/components/CustomTable.tsx
+++ b/frontend/src/components/CustomTable.tsx
@@ -105,12 +105,9 @@ export const css = stylesheet({
     transition: 'margin 0.2s',
   },
   expandedContainer: {
-    borderRadius: 10,
-    boxShadow: '0 1px 2px 0 rgba(60,64,67,0.30), 0 1px 3px 1px rgba(60,64,67,0.15)',
     margin: '16px 2px',
   },
   expandedRow: {
-    borderBottom: '1px solid transparent !important',
     boxSizing: 'border-box',
     height: '40px !important',
   },
@@ -206,6 +203,36 @@ interface CustomTableState {
   tokenList: string[];
 }
 
+interface CustomTableRowProps {
+  row: Row;
+  columns: Column[];
+}
+
+function calculateColumnWidths(columns: Column[]): number[] {
+  const totalFlex = columns.reduce((total, c) => total += (c.flex || 1), 0);
+  return columns.map(c => (c.flex || 1) / totalFlex * 100);
+}
+
+export const CustomTableRow: React.FC<CustomTableRowProps> = (props: CustomTableRowProps) => {
+  const {row, columns} = props;
+  const widths = calculateColumnWidths(columns);
+  return (
+    <React.Fragment>
+      {
+        row.otherFields.map((cell, i) => (
+          <div key={i} style={{width: widths[i] + '%'}} className={css.cell}>
+            {i === 0 && row.error && (
+              <Tooltip title={row.error}><WarningIcon className={css.icon} /></Tooltip>
+            )}
+            {columns[i].customRenderer ?
+              columns[i].customRenderer!({value: cell, id: row.id}) : cell}
+          </div>
+        ))
+      }
+    </React.Fragment>
+  );
+};
+
 export default class CustomTable extends React.Component<CustomTableProps, CustomTableState> {
   private _isMounted = true;
 
@@ -280,8 +307,7 @@ export default class CustomTable extends React.Component<CustomTableProps, Custo
   public render(): JSX.Element {
     const {filterString, pageSize, sortBy, sortOrder} = this.state;
     const numSelected = (this.props.selectedIds || []).length;
-    const totalFlex = this.props.columns.reduce((total, c) => total += (c.flex || 1), 0);
-    const widths = this.props.columns.map(c => (c.flex || 1) / totalFlex * 100);
+    const widths = calculateColumnWidths(this.props.columns);
 
     return (
       <div className={commonCss.pageOverflowHidden}>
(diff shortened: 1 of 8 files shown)


prodonjs:
/hold

This can wait until next week with @rileyjbauer is back. This adds the rollup and expansion of artifacts by their type and name, so that by default, the latest version is shown first.

Screenshot from 2019-07-03 12-56-14

https://kubeflow-v-n00.endpoints.zhenghui-kubeflow-2.cloud.goog/_/metadata/#/artifacts

The alignment for the expanded rows isn't perfect, but it's pretty close without having to substantially tear apart some of the CustomTable code.

prodonjs:
@rileyjbauer welcome back! Take a look at this when you have a chance. Just as an FYI, I'm out for a week starting on Friday if we want to try to get this in before then.

frontend/src/components/CustomTable.tsx
```diff
@@ -206,6 +203,36 @@ interface CustomTableState {
...
+  return (
+    <React.Fragment>
+      {
+        row.otherFields.map((cell, i) => (
+          <div key={i} style={{width: widths[i] + '%'}} className={css.cell}>
+            {i === 0 && row.error && (
+              <Tooltip title={row.error}><WarningIcon className={css.icon} /></Tooltip>
+            )}
+            {columns[i].customRenderer ?
+              columns[i].customRenderer!({value: cell, id: row.id}) : cell}
```
rileyjbauer: is the `!` still needed with the ternary operator?

frontend/src/pages/ArtifactList.tsx
```diff
@@ -184,7 +192,72 @@ class ArtifactList extends Page<{}, PipelineListState> {
...
+  }
+
+  /**
+   * Groups the incoming rows by name and type pushing all but the first row
+   * of each group to the expandedRows Map.
+   * @param rows
+   */
+  private groupRows(rows: Row[]): Row[] {
+    const flattenedRows = rows.reduce((map, r) => {
+      const stringKey = `${r.otherFields[2]}/${r.otherFields[0]}`;
```
rileyjbauer: A comment here indicating what's at these indices would be helpful

frontend/src/pages/ArtifactList.tsx
```diff
@@ -184,7 +192,72 @@ class ArtifactList extends Page<{}, PipelineListState> {
...
+
+    this.setState({expandedRows});
+    return grouped;
+  }
+
+  /**
+   * Toggles the expansion state of a row
+   * @param index
+   */
+  private toggleRowExpand(index: number): void {
```
rileyjbauer: Perhaps it doesn't need to happen in this PR, but I feel like the expand logic in this file should be moved to the `CustomTable` so that it doesn't need to be reimplemented for future list pages

frontend/src/pages/ArtifactList.test.tsx
```diff
@@ -137,6 +163,26 @@ describe('ArtifactList', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('Renders and sorts Artifacts in ascending order by version',
```
rileyjbauer: Possible to add test(s) for row expansion?

frontend/src/components/CustomTable.tsx
prodonjs: Surprisingly, it does. It seems like the compiler should be able to determine that it's being checked, but it doesn't seem to work.

frontend/src/pages/ArtifactList.tsx
prodonjs: Thanks done!

frontend/src/pages/ArtifactList.tsx
prodonjs: I was initially thinking the same thing. I modeled this after what was done in Pipelines. To change it, I think we'd need to lift the expansion state out of the `Row` interface and maintain it within the `CustomTable`. Since `Rows` are passed in as props to `CustomTable` currently, there's no good way of affecting their state without calling back to the component that provided them. It does seem like a good modification for a future PR.

frontend/src/pages/ArtifactList.test.tsx
prodonjs: Done. Tests added for a row with children and one without.

frontend/src/pages/ArtifactList.tsx
rileyjbauer: Yes, I agree. That sounds like a better solution. Okay to do it later

prodonjs:
Yeah dealing with unexpectedly large list results will be a challenge right now. Hopefully once we get pagination support from the server this won't be an issue since we'll be able to set a fixed result limit. We could still have the scenario where everything rolls to the same group, but I guess we can cross that bridge later.

prodonjs:
/hold

Hold until filter behavior is resolved.

prodonjs:
@rileyjbauer I just pushed a change that removes that awkward sort by artifact type and name. I didn't realize that the Map.entries() method emits items in insertion order so since we apply the desired sort on the full result set first, the rolled-up groups will get inserted in the order we want them to be in.

https://kubeflow-v-n00.endpoints.zhenghui-kubeflow-2.cloud.goog/_/metadata/#/artifacts

@zhenghuiwang - To your comment about the *No other rows in group* message. I would like to hide or disable the toggle when no other rows exist as well but it requires a bit more work and I'd like to make sure this is in before I go on vacation starting Friday. I'll create an issue for that.

rileyjbauer:
Thank you for making that change to the sorting. It feels much more intuitive now!

prodonjs:
Closed by #95

PR #93, opened 2019-07-12
check in openapi_client to avoid traceback

### LINKED ISSUE #91: No kfmd/openapi_client
/kind bug

**What steps did you take and what happened:**
During I try to install and run the sdk demo cases, I found no kfmd/openapi_client, more checked, that should be generated by `make swagger-py-client` and ignored in `gitignore`.

```
......
writing dependency_links to kfmd.egg-info/dependency_links.txt
writing top-level names to kfmd.egg-info/top_level.txt
writing manifest file 'kfmd.egg-info/SOURCES.txt'
error: package directory 'kfmd/openapi_client' does not exist
```

Since that's import by kfmd/metadata.py, personally I think we should check in the openapi_client code, at least we should add related reminding to new user to run `make swagger-py-client`, but if so seems it's not whole python pacakge.  so I think we should check in the openapi_client. Thanks.

jinchihe:
While I'm trying to setup `kfmd` and run the sdk demo cases, Traceback happens due to no `kfmd/openapi_client`, more checked, that should be generated by `make swagger-py-client`. I think we should check in so that the sdk can works normally. Thanks.

**Which issue(s) this PR fixes**:
Fixes #91 

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/93)


zhenghuiwang:
/hold

There are end2end tests to generate the client on the fly and test certain scenarios between client/server to prevent regression.

 In the past, various code gen tools were used to do the job. checking in the gen code often results in large unwanted code change, which no human actually reviews.

jinchihe:
Seems the sdk has been installed by `pip install kfmd`.  Thanks @zhenghuiwang 

/close

PR #99, opened 2019-07-20
Update Python library

zhenghuiwang:
Fixes three issues related to python library:

1. The `kfmd.openapi_client` pkg not found in published `kfmd` package. We have to refer the pacakge as `kfmd.openapi_client` instead of `openapi_client` in our source code. (Changes the `make` target for this)

2. For optional properties with `None` value, we need to `del` the key. Otherwise, the server returns 400.

3. Fixes the demo notebook and adds a test for the notebook.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/99)


sdk/python/tests/test.py
```diff
@@ -107,6 +107,27 @@ def test_log_invalid_artifacts_should_fail(self):
     ))
     self.assertRaises(ValueError, e.log_output, artifact2)
 
+  def test_log_metadata_successfully_with_minimum_inforamtion(self):
+    ws1 = metadata.Workspace(backend_url_prefix="127.0.0.1:8080", name="ws_1")
+
+    r = metadata.Run(workspace=ws1, name="first run")
+
+    e = metadata.Execution(name="test execution", workspace=ws1, run=r)
+    assert e.id
```
neuromage: Are we checking that this is not none or true? If so, prefer `self.assertIsNotNone(e.id)` or similar.

Same below

sdk/python/kfmd/metadata.py
```diff
@@ -14,8 +14,8 @@
 
 import datetime
 import json
-import openapi_client
-from openapi_client import Configuration, ApiClient, MetadataServiceApi
+import kfmd.openapi_client as openapi_client
+from kfmd.openapi_client import Configuration, ApiClient, MetadataServiceApi
```
neuromage: style nit: prefer to import modules: 
http://google.github.io/styleguide/pyguide.html#22-imports

sdk/python/tests/test.py
zhenghuiwang: Done!

sdk/python/kfmd/metadata.py
zhenghuiwang: Done.

neuromage:
/lgtm
/approve
Thanks @zhenghuiwang !

PR #104, opened 2019-07-23
Split the dockerfile

zhenghuiwang:

- Split the dockerfile so that different base images can be used.
- Improve efficiency: prebuild a base image so that e2e tests don't have build images from scratch.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/104)


dreamryx:
@zhenghuiwang I will commit the power base dockerfile. The unified method for all kubeflow components will be discussed and I add your method into the doc. The doc is https://docs.google.com/document/d/1CDYYdWi9qP96qaddyI8eO3r1w3yXb5wNBfE5csnRgVs/edit?usp=sharing

PR #110, opened 2019-08-16
Support update artifact/execution type in metadata

dreamryx:
Supporting update artifact/execution type in metadata. For now, the update only support add fields.

Fixes #

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/110)


api/service.proto
```diff
@@ -35,6 +35,24 @@ message CreateArtifactTypeResponse {
   ml_metadata.ArtifactType artifact_type = 1;
 }
 
+message PutTypeOptions {
```
zhenghuiwang: Can we rename PutTypeOptions to `UpdateTypeOptions`? `Put` is too specific to HTTP verb.

But I don't think we need it. See my comment below.

service/service.go
```diff
@@ -158,6 +158,51 @@ func (s *Service) CreateArtifactType(ctx context.Context, req *api.CreateArtifac
...
+
+	if req.ArtifactType == nil {
+		return nil, errors.New("no ArtifactType specified")
+	}
+
+	if req.PutTypeOptions == nil {
+		return nil, errors.New("no PutTypeOptions specified")
+	}
+
+	if req.PutTypeOptions.CanDeleteFields {
```
zhenghuiwang: The only acceptable PutTypeOptions now is

```
putTypeOptions {
					CanAddFields:       true
					CanDeleteFields:    false
					AllFieldsMustMatch: true}
```

How about not exposing PutTypeOptions in the API and add it back later if more possible configs are supported?

service/service_test.go
```diff
@@ -203,6 +203,95 @@ func TestCreateArtifactType(t *testing.T) {
 	}
 }
 
+func TestUpdateArtifactType(t *testing.T) {
```
zhenghuiwang: Can you also write this unit test as integration test? The integration script is at

https://github.com/kubeflow/metadata/blob/master/test/e2e/make_requests.sh

The presubmit  job will start a server and make these requests.

service/service.go
dreamryx: I'm OK with it. We shall expose this later.

service/service_test.go
dreamryx: Yes. Of course. Thanks for the comment.

api/service.proto
dreamryx: We shall expose it later. But the update function does not expose the option looks a bit weird for now.

service/service.go
```diff
@@ -158,6 +158,34 @@ func (s *Service) CreateArtifactType(ctx context.Context, req *api.CreateArtifac
...
+	var putTypeOptions mlmetadata.PutTypeOptions
+
+	if req.ArtifactType == nil {
+		return nil, errors.New("no ArtifactType specified")
+	}
+
+	putTypeOptions.AllFieldsMustMatch = true
+	putTypeOptions.CanAddFields = true
+
+	_, err := s.store.PutArtifactType(req.ArtifactType, &putTypeOptions)
```
zhenghuiwang: It is more readable if you can use struct literal:
```
s.store.PutArtifactType(req.ArtifactType, &putTypeOptions {
   AllFieldsMustMatch : true,
   CanAddFields: true,
})
```

service/service.go
```diff
@@ -310,6 +338,34 @@ func (s *Service) CreateExecutionType(ctx context.Context, req *api.CreateExecut
...
+	var putTypeOptions mlmetadata.PutTypeOptions
+
+	if req.ExecutionType == nil {
+		return nil, errors.New("no ExecutionType specified")
+	}
+
+	putTypeOptions.AllFieldsMustMatch = true
+	putTypeOptions.CanAddFields = true
+
+	_, err := s.store.PutExecutionType(req.ExecutionType, &putTypeOptions)
```
zhenghuiwang: Please use struct literal here.

dreamryx: OK.

service/service.go
dreamryx: OK. No problem.

zhenghuiwang:
/lgtm

@neuromage any comments?

dreamryx:
@zhenghuiwang Thanks for your review! :)

PR #111, opened 2019-08-16
Displays all properties and custom properties in details page

rileyjbauer:
Displays all properties and custom properties in details page instead of only a specific subset.

Also removes expansion button from rows when there aren't multiple versions of an artifact

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/111)

### DIFF
--- a/frontend/src/components/ArtifactInfo.tsx
+++ b/frontend/src/components/ArtifactInfo.tsx
@@ -16,10 +16,8 @@
 import * as React from 'react';
 import {stylesheet} from 'typestyle';
 import {color, commonCss} from '../Css';
-import {formatDateString, getArtifactProperty} from '../lib/Utils';
 import {MlMetadataArtifact} from '../apis/service';
-import {ArtifactProperties, ArtifactCustomProperties} from '../lib/Api';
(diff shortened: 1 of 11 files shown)


frontend/src/components/ArtifactInfo.tsx
```diff
@@ -54,50 +52,35 @@ export class ArtifactInfo extends React.Component<ArtifactInfoProps, {}> {
...
-          </div>
-          {this.props.typeName === 'Model' ?
-            <ModelInfo model={artifact} /> :
-            <div className={css.field}>
-              <dt className={css.term}>URI</dt>
-              <dd className={css.value}>{artifact.uri}</dd>
-            </div>
+          {Object.keys(artifact.properties || {})
+            // TODO: __ALL_META__ is something of a hack, is redundant, and can be ignored
+            .filter(k => k !== '__ALL_META__')
```
prodonjs: *nit:* There's an enum value for `__ALL_META__` and a few of the other magic keys in the Api module if you want to use it. It seemed to have quite a few things packed under it in the first release but hopefully those become normal properties instead of being nested there.

PR #118, opened 2019-08-22
prevent closed loop

### LINKED ISSUE #119: circular reference in pre-defined schema cause server crash
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
When metadata server starts, it first to load pre-defined schemas of artifact and execution stored in json file. However, if those schemas contain circular reference, the metadata server crash with following error:

runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: **stack overflow**

runtime stack:
runtime.throw(0x14020bd, 0xe)
        GOROOT/src/runtime/panic.go:617 +0x72
runtime.newstack()
        GOROOT/src/runtime/stack.go:1041 +0x6f0
runtime.morestack()
        GOROOT/src/runtime/asm_amd64.s:429 +0x8f

goroutine 1 [running]:
runtime.heapBitsSetType(0xc0059cd6b0, 0x30, 0x30, 0x13cf140)
        GOROOT/src/runtime/mbitmap.go:938 +0xa56 fp=0xc0240003b0 sp=0xc0240003a8 pc=0x8663e6
runtime.mallocgc(0x30, 0x13cf140, 0x1, 0x0)
        GOROOT/src/runtime/malloc.go:969 +0x51c fp=0xc024000450 sp=0xc0240003b0 pc=0x85c12c
runtime.newobject(0x13cf140, 0x0)
        GOROOT/src/runtime/malloc.go:1068 +0x38 fp=0xc024000480 sp=0xc024000450 pc=0x85c788
runtime.makemap_small(0x0)
        GOROOT/src/runtime/map.go:295 +0x2d fp=0xc0240004a0 sp=0xc024000480 pc=0x85d39d
github.com/kubeflow/metadata/schemaparser.(*SchemaSet).SimpleProperties(0xc0002448c0, 0xc000309e40, 0x3c, 0x48, 0xc000338be8, 0x1)
        schemaparser/schemaset.go

The simplest way to reproduce this bug is to make the schema quote itself. For example, add `" $ref": "http://github.com/kubeflow/metadata/schema/alpha/artifact.json"` to [artifact.json](http://github.com/kubeflow/metadata/schema/alpha/artifact.json) file.

Currently all the schemas are well designed and tested. But the server need to plug this hole.

dreamryx:
Since the schema contains parent ref, we need to prevent closed loop. Otherwise error like "stack overflow" may comes.

Fixes #119

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/118)

### DIFF
--- a/schemaparser/register.go
+++ b/schemaparser/register.go
@@ -43,7 +43,18 @@ func RegisterSchemas(service *service.Service, schemaRootDir string) ([]string, 
 		return nil, err
 	}
 	var types []string
+
+	inLoopResult, err := ss.CheckClosedLoop()
+	if err != nil {
+		return nil, err
+	}
+
 	for id := range ss.Schemas {
+		if inLoopResult[id] {
+			glog.Warningf("schema is in invalid closed loop. schema $id = %s", id)
+			continue
+		}
+
 		namespace, typename, err := ss.TypeName(id)
 		if err != nil {
 			glog.Warningf("schema has invalid 'kind', 'apiversion', or 'namespace'. schema $id = %s: %v", id, err)
--- a/schemaparser/schemaset.go
+++ b/schemaparser/schemaset.go
@@ -30,7 +30,7 @@ const (
 	versionPropertyName   = "apiversion"
 	idPropertyName        = "id"
 	namePropertyName      = "name"
-	uriPropertyName      = "uri"
+	uriPropertyName       = "uri"
 )
 
 // SimpleProperties are properties of type integer, double and string. The map is from property name to its type.
@@ -49,6 +49,10 @@ type SchemaSet struct {
 	loader  *gojsonschema.SchemaLoader
 	Schemas map[string]*Schema
 }
+
+// CheckResult are the checking result of whether the schema is in a closed loop.
+// The value is true means the schema is in a closed loop
+type CheckResult map[string]bool
 
 // NewSchemaSetFromADir create a SchemaSet from all schema files under a directory and its descendant directories.
 func NewSchemaSetFromADir(dir string) (*SchemaSet, error) {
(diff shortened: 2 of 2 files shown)


zhenghuiwang:
@dreamryx why do you think this PR is needed? Is there a specific bug to be fixed?

The schema registration package is for predefined types and it has unit tests to verify the schemas actually load without errors such as stack overflow.

dreamryx:
@zhenghuiwang Thanks for your review. :) 

It's really a QA test case here:
1. Assume there are three schema a, b and c.
2. Assume a has parent ref b, b has parent ref c and c has parent ref a.
3. When load pre-defined schema, our server may go over like this "a -> b -> c -> a ...."

Yes. As you said, currently the pre-defined schema are well designed and tested.

But we need to add a guard to eliminate the concern in server:
1. User can add schemas in dockerfile to create new image that contains new pre-defined schema.
2. Maybe we could allow user to load their pre-defined schema in the future.

This PR could help improve the server. Hope you can approve it.

Thanks.

zhenghuiwang:
@dreamryx I agree there can be theoretical circular reference problem for the schemas. Is there any specific example of schema that you want to include that has circular reference? 

IMHO we should not have circular reference at the first place and we should let the server panic (due to overflow) during start if there are. Circular reference is not just hard to write code to parse but also hard for human to consume. We should strive to make the schemas self-contain.

dreamryx:
@zhenghuiwang The circular reference is QA test case for now. I do not have an example to use circular reference. 

When there are complicated pre-defined schemas, it is hard for human to check circular reference and find which schema is wrong if circular reference comes up. The server crash with "stack overflow" message can not help user to know what happens and how to fix it quickly. I think it's the server's responsibility to help user check it and report the info to user.

hougangliu:
when server starts, proactive schema check and a clear error message can help indeed instead of panic message.
@dreamryx just one comment. SchemaJSON check should be a common case. So I wonder if has existing related lib to do such check including close-loop check instead of implementating it by yourself here.

Issue #120, opened 2019-08-27
How to make original ArtifactType 


TomomasaTakatori:
In python SDK, kinds of ArtifactType are only Dataset, Model, and Metrics.
If another ArtifactTypes is needed, we can't make them (We could can do by using module generated by open_api, but there is no document ).
And kfmd's concepts is slightly differed from original MLMD library. For example, ExecutionType is not able to be define.

Summarization
- Easily making original ArtifactType (If currently can, please tell me how to do it ) .
- Following MLMD library.

dreamryx:
Assume **ws1** is your workspace create by "metadata.Workspace". You can create an ArtifactType named **tomoma_type** with tree fileds as below:

```
artifactType = openapi_client.MlMetadataArtifactType(
    name="kubeflow.org/alpha/tomoma_type",
    properties={
        "owner":"STRING",
        "create_time":"STRING",
        "description":"STRING",
    })  
response = ws1.client.create_artifact_type(artifactType)
```

I agree that the SDK should be more convenient to use.  I want to update the sdk later.

TomomasaTakatori:
Thank you for your replying.

I understand If we would like to create ArtifactType, we should write original code.

And the type of ExecutionType is just "kubeflow.org/alpha/execution". 
Is it the same as ArtifactType to create ExecutionType, that mean we need to write original code ?

dreamryx:
Yes. If you want to create new executionType, you need to write the similar code showed above. The sdk python lib is only for the pre-defined type and objects. :)

TomomasaTakatori:
Thank you so much.

PR #122, opened 2019-09-02
improve the necessary input arguments checking for python

dreamryx:
Add some code for checking the necessary input arguments in python sdk.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/122)

### DIFF
--- a/sdk/python/kfmd/metadata.py
+++ b/sdk/python/kfmd/metadata.py
@@ -50,8 +50,6 @@ class Workspace(object):
           raise ValueError("'backend_url_prefix' must be set and in string type.")
     if name is None and type(name) != str:
           raise ValueError("'name' must be set and in string type.")
-    if labels is None:
-          raise ValueError("'labels' must be set.")
     self.backend_url_prefix = backend_url_prefix
     self.name = name
     self.description = description
--- a/sdk/python/kfmd/metadata.py
+++ b/sdk/python/kfmd/metadata.py
@@ -45,10 +45,9 @@ class Workspace(object):
       description {str} -- Optional string for description of the workspace.
       labels {object} Optional key/value string pairs to label the workspace.
     """
-    # TODO(zhenghuiwang): check each field's type and whether set.
-    if backend_url_prefix is None and type(backend_url_prefix) != str:
+    if backend_url_prefix is None or type(backend_url_prefix) != str:
           raise ValueError("'backend_url_prefix' must be set and in string type.")
-    if name is None and type(name) != str:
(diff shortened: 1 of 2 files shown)


dreamryx:
@zhenghuiwang Please help review it. Thanks. :) 
By the way, do you think there is a need to write some wrapper api in metadata.py to help user create artifactType and so on easily. Please refer to #120.

sdk/python/kfmd/metadata.py
```diff
@@ -46,6 +46,10 @@ def __init__(self,
       labels {object} Optional key/value string pairs to label the workspace.
     """
     # TODO(zhenghuiwang): check each field's type and whether set.
```
zhenghuiwang: Can you remove the this TODO?

Same below.

sdk/python/kfmd/metadata.py
```diff
@@ -46,6 +46,10 @@ def __init__(self,
...
+    if backend_url_prefix is None and type(backend_url_prefix) != str:
+          raise ValueError("'backend_url_prefix' must be set and in string type.")
+    if name is None and type(name) != str:
```
zhenghuiwang: it should be `if name is None or type(name) != str`, not `and`.

Same below.

sdk/python/kfmd/metadata.py
```diff
@@ -46,6 +46,10 @@ def __init__(self,
       labels {object} Optional key/value string pairs to label the workspace.
     """
     # TODO(zhenghuiwang): check each field's type and whether set.
+    if backend_url_prefix is None and type(backend_url_prefix) != str:
```
zhenghuiwang: Could you add some unit tests for this?

sdk/python/kfmd/metadata.py
dreamryx: Yes. I will change it now.

sdk/python/kfmd/metadata.py
dreamryx: Yes. Of course.

sdk/python/kfmd/metadata.py
dreamryx: Thanks for your review!

sdk/python/kfmd/metadata.py
dreamryx: OK. Got it.

zhenghuiwang:
Thanks!

/lgtm
/approve

dreamryx:
@zhenghuiwang   Thanks for your review!

PR #137, opened 2019-10-15
support creating executions and atom execution with related artifacts and events

dreamryx:
This PR help support creating executions and atom execution with related artifacts and events for user.  The new function support user to create new execution, new related artifacts and new related events at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/137)


dreamryx:
@zhenghuiwang  Please help review this PR. Thanks. : )

zhenghuiwang:
@dreamryx Thanks for sending this PR!

I see similar functionality in the gRPC service of MLMD.
https://github.com/google/ml-metadata/blob/master/ml_metadata/proto/metadata_store_service.proto#L120

Is this PR intended to make KFMD HTTP server expose the MLMD gRPC method?

In the long run, it is better to deploy MLMD gRPC + envoy HTTP proxy for the gRPC service. So that we don't need to keep updating this HTTP server for newer functionalities in MLMD gRPC. To gracefully transit to the MLMD gRPC + envoy, it would be nice to keep KFMD HTTP server API unchanged for the 0.7 release and deprecate it in Kubeflow 1.0. 

I think you can start use KFMD gRPC service as it will be available in v0.7.
https://github.com/kubeflow/manifests/blob/master/metadata/base/metadata-deployment.yaml#L40

dreamryx:
@zhenghuiwang OK. Got it.  Thanks!

Issue #155, opened 2019-11-04
Update the python SDK to talk to MLMD gRPC service


zhenghuiwang:
/kind feature

**Describe the solution you'd like**
Current Python SDK talks to the backend HTTP metadata service, which is built on top of MLMD library. For MLMD >= 0.15.0, it comes with a gRPC service and a Python connection option to talk to the gRPC service.

Kubeflow >= 0.7 deploys MLMD gRPC service. If the python SDK can talk to the MLMD gRPC service, then we don't need to keep the HTTP service up to date with MLMD library.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

zhenghuiwang:
Fixed in https://github.com/kubeflow/metadata/pull/156

PR #156, opened 2019-11-04
Update Python SDK to use MLMD gRPC service

zhenghuiwang:
Related to https://github.com/kubeflow/metadata/issues/155

This PR updates the python SDK.
- New `Store` class is introduced for connecting to MLMD gRPC service, which returns `ml_metadata.metadata_store.MetadataStore`.
- Old `backend_url` option for connecting to HTTP service is deprecated.
- Change E2E test to test python SDK with gRPC service.

It is a breaking change, since the way to connect backend service changes. I keep the `backend_url` as optional parameter to ask users to update their code if the `backend_url` is set.

I will update the demo notebook and upload the package to pypi in next PR.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/156)

### DIFF
--- a/sdk/python/kubeflow/metadata/metadata.py
+++ b/sdk/python/kubeflow/metadata/metadata.py
@@ -78,12 +78,12 @@ class Workspace(object):
       backend_url_prefix {str} -- Deprecated. Please use 'store' parameter.
     """
     if backend_url_prefix:
-          raise ValueError("""'backend_url_prefix' is deprecated. Please set
-          Metadata.Store parameter to connect to the metadata gRPC service.""")
+      raise ValueError("""'backend_url_prefix' is deprecated. Please set
+        Metadata.Store parameter to connect to the metadata gRPC service.""")
     if name is None or type(name) != str:
-          raise ValueError("'name' must be set and in string type.")
+      raise ValueError("'name' must be set and in string type.")
     if not store or type(store) != Store:
-          raise ValueError("'store' must be set as metadata.Store")
+      raise ValueError("'store' must be set as metadata.Store")
     self.store = store.store
     self.name = name
     self.description = description
@@ -113,10 +113,10 @@ class Workspace(object):
       "id": artifact.id,
     }
     if artifact.custom_properties is not None:
-        if WORKSPACE_PROPERTY_NAME in artifact.custom_properties:
-          result["workspace"] = artifact.custom_properties[WORKSPACE_PROPERTY_NAME].string_value
-        if RUN_PROPERTY_NAME in artifact.custom_properties:
-          result["run"] = artifact.custom_properties[RUN_PROPERTY_NAME].string_value
+      if WORKSPACE_PROPERTY_NAME in artifact.custom_properties:
+        result["workspace"] = artifact.custom_properties[WORKSPACE_PROPERTY_NAME].string_value
+      if RUN_PROPERTY_NAME in artifact.custom_properties:
+        result["run"] = artifact.custom_properties[RUN_PROPERTY_NAME].string_value
     if not artifact.properties:
       return result
     for k,v in artifact.properties.items():
@@ -149,9 +149,9 @@ class Run(object):
       description {str} -- Optional description.
     """
     if workspace is None:
-          raise ValueError("'workspace' must be set.")
+        raise ValueError("'workspace' must be set.")
     if name is None or type(name) != str:
-          raise ValueError("'name' must be set and in string type.")
+        raise ValueError("'name' must be set and in string type.")
     self.workspace = workspace
     self.name = name
     self.description = description
@@ -177,9 +177,9 @@ class Execution(object):
     Returns an execution object for logging.
     """
     if workspace is None:
-          raise ValueError("'workspace' must be set.")
+        raise ValueError("'workspace' must be set.")
     if name is None or type(name) != str:
-          raise ValueError("'name' must be set and in string type.")
+        raise ValueError("'name' must be set and in string type.")
     self.id = None
     self.name = name
     self.workspace = workspace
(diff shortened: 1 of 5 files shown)


sdk/python/kubeflow/metadata/metadata.py
```diff
@@ -26,38 +27,67 @@
...
+
 class Workspace(object):
   """
   Groups a set of runs of pipelines, notebooks and their related artifacts
   and executions.
   """
 
   def __init__(self,
-               backend_url_prefix=None,
+               store= None,
```
jinchihe: mini: extra space

sdk/python/kubeflow/metadata/metadata.py
```diff
@@ -26,38 +27,67 @@
...
-                                  backend, e.g. "127.0.0.1:8080".
+      backend_url_prefix {str} -- deprecated.
       name {str} -- Required name for the workspace.
       description {str} -- Optional string for description of the workspace.
       labels {object} Optional key/value string pairs to label the workspace.
     """
-    if backend_url_prefix is None or type(backend_url_prefix) != str:
-          raise ValueError("'backend_url_prefix' must be set and in string type.")
+    if backend_url_prefix:
+          raise ValueError("""'backend_url_prefix' is deprecated. Please set
```
jinchihe: is that better to WARNING here and remove from next release?

test/scripts/setup-services-run-tests.sh
```diff
@@ -145,6 +143,10 @@ kubectl -n $NAMESPACE port-forward $TARGET_GRPC_POD 8081:8080 &
 # Stream server logs.
 kubectl -n $NAMESPACE logs -f $TARGET_GRPC_POD &
 
+# Run Python tests
+pip install ml-metadata==0.15.0
```
jinchihe: Is that better ">=" to be consistent with REQUIRES so catch the problem if new ml-metadata introduces?

zhenghuiwang: Good suggestion. I change setup.py to use `==`.

sdk/python/kubeflow/metadata/metadata.py
zhenghuiwang: My original plan is to do what you suggested, but realized it would require to keep two ways of calling backends and the code would look messy.

Whoever want to use the previous SDK should ping down the version.

zhenghuiwang:
Thanks for the review @jinchihe 

I updated the PR and it is ready for review @gaoning777

sdk/python/kubeflow/metadata/metadata.py
```diff
@@ -14,294 +14,335 @@
...
+      grpc_host: Required gRPC service port.
+      root_certificates: Optional SSL certificate for secure connection.
+      private_key: Optional private_key for secure connection.
+      certificate_chain: Optional certificate_chain for secure connection.
+    """
+    config = mlpb.MetadataStoreClientConfig()
+    config.host = grpc_host
+    config.port = grpc_port
+    if root_certificates or private_key or certificate_chain:
+      config.ssl_config = config.SSLConfig()
```
gaoning777: Should this be mlpb.MetadataStoreClientConfig.SSLConfig()? Maybe Protobuf support referencing the class through the instance.

sdk/python/kubeflow/metadata/metadata.py
```diff
@@ -14,294 +14,335 @@
...
+_WORKSPACE_PROPERTY_NAME = '__kf_workspace__'
+_RUN_PROPERTY_NAME = '__kf_run__'
+_ALL_META_PROPERTY_NAME = '__ALL_META__'
+
+
+class Store(object):
+  """Metadata Store that connects to the Metadata gRPC service."""
+
+  def __init__(self,
+               grpc_host: str = "metadata-grpc-service.kubeflow",
```
gaoning777: Maybe the default value be the one that works for the default istio kubeflow deployment, such as metadata-grpc-service.kubeflow.svc.cluster.local? Not sure if my example host is exactly correct.

sdk/python/kubeflow/metadata/metadata.py
zhenghuiwang: It turns out this SubMessage ssl_config field is automatically initialized with properties with None value. So I removed this line.

But I found bug of MLMD when writing unit test:

```
 File "/Users/zhenghui/kubeflow/metadata/.env/lib/python3.6/site-packages/ml_metadata/metadata_store/metadata_store.py", line 103, in _get_channel
    certificate_chain)
  File "/Users/zhenghui/kubeflow/metadata/.env/lib/python3.6/site-packages/grpc/__init__.py", line 1593, in ssl_channel_credentials
    certificate_chain))
  File "src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi", line 133, in grpc._cython.cygrpc.SSLChannelCredentials.__cinit__
TypeError: expected certificate to be bytes, got <class 'str'>
```

gRPC requires all three fields to be bytes (https://grpc.github.io/grpc/python/grpc.html#grpc.ssl_channel_credentials)

But MLMD pass them in as string.

The fix is outside the scope of this PR.

sdk/python/kubeflow/metadata/metadata.py
zhenghuiwang: `<service>.<namespace>` should work as DNS inside a k8s cluster. I've used this pattern in multiple notebooks running in KF.

zhenghuiwang:
Thanks @gaoning777 @jinchihe for review!

@gaoning777  addressed your comments. PTAL

sdk/python/kubeflow/metadata/metadata.py
gaoning777: If it works, great. I saw the other file also uses the <service>.<namespace>(https://github.com/kubeflow/metadata/blob/master/watcher/dockerfiles/watcher.yaml#L21).

I've read k8s doc that the DNS resolves the service as my-svc.my-namespace.svc.cluster-domain.example(https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-records). Do you know if <service>.<namespace> is resolved by Istio DNS or something else? Happy to learn it.

zhenghuiwang: `<service>.<NS>` is a way how k8s resolves DNS. Please see
https://kubernetes.io/docs/concepts/services-networking/service/#dns

gaoning777: Thanks

sdk/python/kubeflow/metadata/metadata.py
gaoning777: Would you mind creating a bug for this?

zhenghuiwang: https://github.com/google/ml-metadata/issues/22

gaoning777: Thanks

hughmiao: @gaoning777 , shall we address this along with the python client grpc server test?

gaoning777: SGTM.

Issue #138, opened 2019-10-15
Multiple CrashLoopBackoffs related to db migration


yanniszark:
/kind bug

**What steps did you take and what happened:**
Deploy Kubeflow using the metadata package from manifests.
The image tag is: `v0.1.9`.

20 minutes after deploying, all servers  were in CrashLoopBackoff.
The logs seemed related to a migration:

```text
F1015 10:54:29.242102       1 main.go:90] Failed to create ML Metadata Store with config mysql:<host:"metadata-db.kubeflow" port:3306 database:"metadb" user:"root" password:"test" > : mysql_query failed: errno: 1054, error: Unknown column 'is_artifact_type' in '
Type'
        Failed to migrate existing db; the migration transaction rolls back..
goroutine 1 [running]:
github.com/golang/glog.stacks(0xc0001ec300, 0xc0001f6000, 0x153, 0x1a9)
        external/com_github_golang_glog/glog.go:769 +0xb1
github.com/golang/glog.(*loggingT).output(0x16ba6c0, 0xc000000003, 0xc000032070, 0x156222b, 0x7, 0x5a, 0x0)
        external/com_github_golang_glog/glog.go:720 +0x2f6
github.com/golang/glog.(*loggingT).printf(0x16ba6c0, 0x3, 0xfd0bfa, 0x37, 0xc0001bde30, 0x2, 0x2)
        external/com_github_golang_glog/glog.go:655 +0x14e
github.com/golang/glog.Fatalf(...)
        external/com_github_golang_glog/glog.go:1148
main.mlmdStoreOrDie(0x0)
        server/main.go:90 +0x1c3
main.main()
        server/main.go:101 +0xe0
```

Issue remains even after deleting and recreating all the Pods.

**What did you expect to happen:**
I expected Pods to become Ready without any restarts.

**Environment:**

- Metadata version: v0.1.9
- Kubeflow version: v0.7.0-rc1
- Kubernetes version: (use `kubectl version`): 1.14
- OS (e.g. from `/etc/os-release`): Ubuntu

zhenghuiwang:
/cc @dushyanthsc @neuromage 

I think it is related to the MLMD upgrade in https://github.com/kubeflow/metadata/pull/130

@yanniszark  does the database `metadb` exist before this installation or not?

yanniszark:
@zhenghuiwang it was a brand new GKE cluster.
Nothing should have been there.

/priority p0
/cc @jlewi

[...]

PR #157, opened 2019-11-05
Add retry when connecting metadata store

zhenghuiwang:
related to #138 

Add retries when failed to connected to MLMD store. `i`th retry will wait ~`2^(i-1)` seconds.

Tested it on a KF cluster and the `metadata-service` has 0 crash of 2 replicas.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/157)

### DIFF
--- a/server/main.go
+++ b/server/main.go
@@ -19,8 +19,11 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
+	"math/rand"
 	"net"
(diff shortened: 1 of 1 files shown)


jeffgriffith:
is there an easy workaround for this or just wait for v0.7-rc9? a fresh rc8 cluster never recovers for me (285 restarts)

PR #160, opened 2019-11-07
Replace MLMD HTTP API requests with gRPC-Web requests to ml_metadata

### LINKED ISSUE #148: [MLMD Lineage] Add gRPC Web Client for fetching data from MLMD from gRPC Backend
**Describe the solution you'd like**

Use generated classes for service and API requests in Artifact List and Artifact Details tabs.

**Anything else you would like to add:**

See https://github.com/kubeflow/pipelines/pull/2057 for similar implementation.

/kind feature
/area front-end
/area metadata
/assign @kwasi
/cc @avdaredevil

kwasi:
## Changes

* Add `Api.metadataStoreService` client to request data from `/ml_metadata`
* Proxy all calls to `/ml_metadata*` on metadata-ui frontend sever to `metadata-envoy-proxy` service
* Convert existing calls for Artifacts and Executions to `ml_metadata` calls over gRPC-web 
* Add `npm run start:proxy` to proxy to port 3000 of `metadata-ui` pod. 
  * NOTE: Proxying to port 8080 of `metadata-service` won't work for routing gRPC calls. 

## Dev Changes

* Add mock for @improbable-eng/grpc-web 
* Bump `react-scripts` to 3.2.0 to support `EXTEND_ESLINT` for ignoring generated proto classes
* Add `npm run api:metadata` for generating `grpc-web` protos
* Move jest `__mocks__` into `src/` to work around regression in react-scripts 3.2.0 (https://github.com/facebook/create-react-app/issues/7539)

Fixes #148

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/160)


frontend/src/pages/ExecutionDetails.tsx
```diff
@@ -81,21 +82,34 @@ export default class ExecutionDetails extends Page<{}, ExecutionDetailsState> {
...
+
+    const execution = response!.getExecutionsList()[0];
+    
+    const executionName = getResourceProperty(execution, ExecutionProperties.NAME);
+    let title = executionName ? executionName.toString() : '';
+
+    this.props.updateToolbar({
+      pageTitle: title
+    });
+    this.setState({ execution });
```
avdaredevil: _nit_
```suggestion
    this.setState({execution});
```

frontend/start-proxy.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
avdaredevil: Might be a pain, but could we have this written in Node? (so it's cross-platform, and this bash file can run an npm script to actually work)?

frontend/src/lib/Api.ts
```diff
@@ -37,6 +41,35 @@ export interface ListRequest {
...
+      resolve({ response, error });
+    };
+    apiMethod(request, handler);
+  });
+}
+
+const metadataServiceClient = new MetadataStoreServiceClient('');
+
+// TODO: add all other api methods we need here.
+const metadataServicePromiseClient = {
```
Bobgy: This is a workaround I made, I think the best approach is to use https://github.com/kubeflow/pipelines/issues/2196.

Migration could be a bad experience when a lot more new code depends on it, so I prefer we do this early. What do you think?

kwasi: Sounds good. I had to switch to grpc-web instead of using @improbable-eng/grpc-web.

frontend/start-proxy.sh
kwasi: Done. Also rewrote gen_grpc_web_protos in Node.

frontend/src/pages/ExecutionDetails.tsx
kwasi: Done

kwasi:
Here's a screencast showing the network loads in action: 
https://screencast.googleplex.com/cast/NTczOTE4MTYwMDkzMTg0MHw2ODhjMDk0Yy02MQ

Also pushed a docker image to my cluster:
https://kubeflow-ui-07e.endpoints.kwasinti-kubeflow-dev.cloud.goog/

frontend/src/lib/Api.ts
Bobgy: SG. Is it because the client with promise support is using grpc web client?

kwasi:
Closing this PR as obsolete in favor of #165.

PR #164, opened 2019-11-12
Add arm64 support for metadata

MrXinWang:
Referencing #96 and [kubeflow/issue#2237](https://github.com/kubeflow/kubeflow/issues/2337), add arm64 support for metadata. Have tested the docker image building on both arm64 and x86_64 machines.

Signed-off-by: Henry Wang <<henry.wang@arm.com>>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/164)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,10 @@ mlmd-proto:
 metadata-docker-image:
 	docker build -t gcr.io/kubeflow-images-public/metadata .
 
+metadata-docker-image-aarch64:
+	docker build -t metadata-base-aarch64:latest -f ./dockerfiles/Dockerfile-linux-base.aarch64 . && \
+	docker build -t kubeflow-images-public/metadata --build-arg BASE_IMG=metadata-base-aarch64:latest --build-arg OUTPUT_DIR=linux_arm64_stripped .
+
 swagger-py-client:
 	mkdir -p /tmp/swagger
 	wget http://central.maven.org/maven2/org/openapitools/openapi-generator-cli/4.0.1/openapi-generator-cli-4.0.1.jar -O /tmp/swagger/swagger-codegen-cli.jar && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -26,7 +26,12 @@ mlmd-proto:
 	mv ml_metadata/proto/*.go ml_metadata/
 
 metadata-docker-image:
-	docker build -t gcr.io/kubeflow-images-public/metadata .
+	@if test $(shell uname -m) = "aarch64"; then \
```
zhenghuiwang: Can you add a new build target like `docker-image-aarch64`? The build target doesn't have to tie to build system info.

Makefile
```diff
@@ -26,7 +26,12 @@ mlmd-proto:
...
+		docker build -t metadata-base-aarch64:latest -f ./dockerfiles/Dockerfile-linux-base.aarch64 . && \
+		docker build -t gcr.io/kubeflow-images-public/metadata --build-arg BASE_IMG=metadata-base-aarch64:latest --build-arg OUTPUT_DIR=linux_arm64_stripped .;\
```
zhenghuiwang: Do you have your own docker registry for these images?

`gcr.io` has its own policy about published images but Kubeflow testing infra is on GCP right now which can't test aarch64. Reinforcing `gcr.io` policy on these images will be hard.

You can add links to your registry in the README file if you want.

Makefile
MrXinWang: Yes for sure, and I have updated the commit :)

Makefile
MrXinWang: Hi @zhenghuiwang ! Thanks for your kindly suggestions! I am not sure if Arm has its official docker registry and I will confirm with my colleagues now. I am thinking if I use my own docker registry, it would be unprofessional. Therefore I changed all tags for these images to use local registry, and hopefully you are OK with it. If not I will change again. Thanks for your patience.

zhenghuiwang: Looks good. Thanks!

PR #167, opened 2019-11-14
Python SDK improvements

zhenghuiwang:
This PR improves current Python SDK:
- Adds retries to all outbound gRPC calls.
- Handles the case of logging the same artifact multiple times:
  - Case 1: the artifact already has a ID assigned. It means this artifact was saved before, then no need to save the data again.
  - Case 2: new artifact doesn't have ID, but can be considered as duplication of some artifact in the DB. For example, a model with save `name`, `version` and `uri`. An artifact class is asked to have a optional `is_duplicated` method to determine the case.
- Introduces python ABC for abstract Artifact, which formally states the protocol used before.
- add `__repr__` to better print out artifact objects.

Testing:
- More tests are added.
- Rewrite the test setup to start tests from a virtual env.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/167)

### DIFF
--- a/sdk/python/kubeflow/metadata/metadata.py
+++ b/sdk/python/kubeflow/metadata/metadata.py
@@ -43,7 +43,7 @@ class Artifact(ABC):
     pass
 
   @staticmethod
-  def isDuplicated(a: mlpb.Artifact, b: mlpb.Artifact):
+  def is_duplicated(a: mlpb.Artifact, b: mlpb.Artifact):
     '''Checks if two artifacts are duplicated.
 
     The artifacts may be considered duplication even if not all the fields are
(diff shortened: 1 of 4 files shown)


zhenghuiwang:
/assign @jinchihe @gaoning777 

It is ready for your review. Thanks

jinchihe:
Would please update new version in setup.py file so that we can release new pkg in Pypi for the two new big enhancements? Others look good to me.

zhenghuiwang:
> Would please update new version in setup.py file so that we can release new pkg in Pypi for the two new big enhancements? Others look good to me.

Thanks @jinchihe . `setup.py` is already 0.3.0. (PyPI has 0.2.0) So no need to change it.

I can publish 0.3.0 once this PR is landed.

jinchihe:
@zhenghuiwang Great.
/lgtm

PR #168, opened 2019-11-15
Add sphinx docs for Python SDK.

zhenghuiwang:
Related to https://github.com/kubeflow/website/issues/1206

This PR aims to check in the configuration to auto generate python docs from python docstrings, using readthedocs.io

The improvement of the actual doc content should be done in subsequent PR on docstrings.

Preview: https://kf-metadata-pr-168.readthedocs.io/en/latest/index.html

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/168)


jinchihe:
Hello @zhenghuiwang that's really great to generate docs for SDK, I just done that for kubeflow fairing, may I suggest to update to `sphinx_rtd_theme` style that seems bette, and shall we add content in the index page so that user can click and read that directly as below link:
https://kubeflow-fairing.readthedocs.io/en/latest/index.html

But that's only suggestion, nice to have. That's good for current behavior...haha

zhenghuiwang:
@jinchihe Thanks for the suggestions! I change the theme and it looks much better.

Fairing has many submodules and TOC helps to navigate among them. This sdk has a single module with a bunch of classes. I don't find a good sphinx plugin to automatically generate the TOC for these classes. I will leave the navigation to the `Index` page or search.

jinchihe:
Sounds great!
/lgtm

sarahmaddox:
Thanks @zhenghuiwang! When the SDK docs are available on readthedocs.io, we can add a link to them from the Kubeflow website. I'll add a comment to the issue.

/lgtm

PR #178, opened 2019-12-02
Add a MLMD context for every `workspace` & update notebook example

zhenghuiwang:
`workspace` is for group relevant metadata together. MLMD recently added generic [Context&ContextType](https://github.com/google/ml-metadata/blob/master/g3doc/get_started.md#concepts) for the same purpose. 

This PR adds one context for each workspace to enable fast metadata retrieval via `context`. Current way of adding workspace name as artifact property shall be gradually phased out in the future.(Current UI still needs it)
- Context name is required to be unique for different contexts. This PR adds the option of whether reusing context with same name when initializing the workspace.

Changes:

- Make python tests idempotent. Update notebook test setup to use SDK HEAD.

- Update `demo.ipynb` notebook with the latest SDK.

- Update README, setup.py and `__init__` to prepare SDK release.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/178)


zhenghuiwang:
@jinchihe and @gaoning777 This PR is ready for your review! thanks

sdk/python/README.md
```diff
@@ -9,13 +9,17 @@ You can install this Python library for logging metadata via the following comma
 pip install kubeflow.metadata
 ```
 
+## Releases
+- 0.3.0 to be released with Kubeflow v0.1. ([doc](https://master.kubeflow.org/docs/components/misc/metadata/), [tutorial](https://github.com/kubeflow/metadata/blob/master/sdk/python/sample/demo.ipynb), [SDK Reference](https://kubeflow-metadata.readthedocs.io/en/latest/))
```
jinchihe: Kubelow 0.1 -> 0.7?

sdk/python/kubeflow/metadata/metadata.py
```diff
@@ -176,6 +185,44 @@ def _flat(self, artifact):
         result[k] = v
     return result
 
+  def _get_context_id(self, reuse_workspace_if_exists):
+    ctx = self._get_existing_context()
+    if ctx is not None:
+      if reuse_workspace_if_exists:
+        return ctx.id
+      else:
+        raise ValueError('workspace name {} already exists with id {}'.format(
```
jinchihe: Suggest making more clear reminder here such as set `reuse_workspace_if_exists=True` if want to reuse workspace ...

sdk/python/sample/demo.ipynb
```diff
@@ -495,9 +326,9 @@
...
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.7"
+   "version": "3.6.0"
   }
  },
  "nbformat": 4,
  "nbformat_minor": 2
-}
+}
```
jinchihe: nit: add new line.

sdk/python/setup.py
```diff
@@ -26,9 +26,10 @@
     name=NAME,
     version=VERSION,
     description='KubeFlow Metadata SDK',
+    long_description='Please visit the document page: https://github.com/kubeflow/metadata/tree/master/sdk/python',
```
jinchihe: since below url points to the github repo, do you think that's better to piont to readthedos url for the sdk?

sdk/python/tests/run_tests.sh
```diff
@@ -13,13 +13,19 @@
...
 python3 -m pip install pytest
 # install local kubeflow.metadata package
 python3 -m pip install -e .
-python3 -m pytest tests/test.py
-rm -rf .testing-env
+python3 -m pytest ./tests
+# Run integration tests multiple times should get the same result. This checks
+# the idenpotence of the tests.
+echo "Run tests the second time:"
+python3 -m pytest ./tests
```
jinchihe: sorry, confused me, how to check the idenpotence for the twice running here? or you want to test reuse case if existing, no need check, just ensure case pass?

test/scripts/setup-services-run-tests.sh
```diff
@@ -125,28 +131,37 @@ until curl -H "ContentType: application/json" localhost:8080/api/v1alpha1/artifa
...
 
+# Test demo notebook
+pip3 install jupyterlab
+pip3 install nbconvert
+pip3 install pandas
+# Use local server and package.
+sed -i -e "s@metadata-grpc-service.kubeflow@127.0.0.1@" sample/demo.ipynb && \
+sed -i -e "s@grpc_port=8080@grpc_port=8081@" sample/demo.ipynb && \
+sed -i -e "s@pip install kubeflow-metadata --user@pip install -e ${SRC_DIR}/sdk/python@" sample/demo.ipynb && \
+  python3 -m nbconvert --to notebook --execute sample/demo.ipynb
```
jinchihe: Seems we cannot ensure the notebook output is consistent with expected? May be better use `papermill.execute_notebook`, we can set the parameters directly without `sed -i` and can assert the outputs.

sdk/python/tests/run_tests.sh
zhenghuiwang: Context: one test here is almost the same as the sample notebook. It is ideal that multiple notebook runs continue to work.

Sorry, idempotency is not the right term here, but these tests now update and depend on DB states. Just want to make sure each individual test continue working even if the DB has previous run state.

test/scripts/setup-services-run-tests.sh
zhenghuiwang: Thanks for pointing papermill. I think it is nice tool to automate the tests with the bash hacks.

You are right we don't check the output string now and it can be improved with papermill. I leave a TODO and want to do it in next PR.

sdk/python/tests/run_tests.sh
jinchihe: Oh.. Make sense then, thanks @zhenghuiwang

jinchihe:
@zhenghuiwang Thanks. Sounds great!

/lgtm

zhenghuiwang:
Thanks for the review!
/approve

PR #182, opened 2019-12-07
Add third party licenses into docker image

zhenghuiwang:
To comply with the license requirements, such as BSD-3 or MIT, of third party dependencies, we need to copy all license contents into the docker images.

This PR adds instructions&scripts to gather these license information.

Related to: kubeflow/testing#539

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/182)


third_party_licenses/README.md
```diff
@@ -0,0 +1,100 @@
...
+    https://github.com/ghodss/yaml/blob/master/LICENSE,MIT
+    https://github.com/gogo/protobuf/blob/master/LICENSE,BSD 3-Clause "New" or "Revised" License
+    ......
+    ```
+
+    Finally, we can patch the additional license types in `additional_license_info.csv` on `license_info.csv` to get the final list of licenses with types.
+
+    ```
+    $  python patch_additional_license_info.py
+    ```
```
Bobgy: Thank you! This is awesome!
Workflow for these is exactly something I didn't have time to polish.

Can I copy this script back to kfp repo? (I'm not sure if there would be legal issues.)

I think it would be the easiest for up-comers to find all relevant tools in one place. What extra efforts do we need to take if we all move these into a dedicated repo? (asking just because I'm not familiar with usual process)

third_party_licenses/README.md
```diff
@@ -0,0 +1,100 @@
+This folder contains a list of license file locations and types in `license_info.csv`  for all dependencies and transitive dependencies of the metadata docker image.
+
+The actual license contents of all dependencies shall be generated via the following command and put into the docker image:
+```
+$ python concatenate_license.py
```
Bobgy: I'm curious why you decided to maintain the list and download them when building the image.
Are there any concerns committing concatenated license file into the repo?

I this approach because I worry about flakiness for image building. (Maybe it's not as bad as I thought, because github is stabler than other websites.)

third_party_licenses/README.md
jlewi: @Bobgy and @zhenghuiwang 
We use kubeflow/testing as our engprod repository. If we are building tooling that can be reused across applications for satisfying licensing requirements (I hope we are). Then I would suggest putting that tooling in that repository.

@Bobgy does that work for you?

third_party_licenses/README.md
zhenghuiwang: My concern is the correctness of the final license file.

The final license txt file is very large as I have seen it locally. I think mistakes of editing the file can happen if hundreds of lines are changed. If we check in the file, we will rely on the person do it right locally.

Having said that, flakiness of this script is also a concern if it happens quite often. I would leave this PR as it is now. If it makes the building process unstable, I can commit the final file and add a test to `diff` the script output with the final file.

third_party_licenses/README.md
zhenghuiwang: Moving all scripts of image building process into one place like kubeflow/testing SGTM.

@Bobgy If you agree, you could copy your cli to a folder under kubeflow/testing and I can update it.

Bobgy: sounds good to me, I will make a PR

third_party_licenses/README.md
Bobgy: SG, thanks for explaining your reasoning behind

zhenghuiwang:
> Presubmit check looks good. It seems you should also exit 1 here. Did I miss anything?

You're right. Done. Thanks

jlewi:
/lgtm
/approve

/hold 
for @Bobgy

PR #183, opened 2019-12-08
Reduce EdgeCanvas width to 120px and parameterize card and edge width 

kwasi:
/area metadata
/area front-end
/kind feature
/assign @avdaredevil 

## Changes

* EdgeCanvas width is reduced by 80px total (requested by UX)

Before: 
* https://screenshot.googleplex.com/WmcEciDEVF8
* https://screenshot.googleplex.com/3aZJr9ba1zM (200px edge canvas)

After: 

* https://screenshot.googleplex.com/NMbCnqry2ea 
* https://screenshot.googleplex.com/uHViiuPn7qE (120px edge canvas)

## Implementation Changes

* `<EdgeCanvas />` and `<LineageCard />` width are controlled by constants in `LineageCss.ts`
* Column width padding is now based on the `<EdgeCanvas />` width
* Column width is now equal to the card width

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/183)

PR #185, opened 2019-12-09
Move LineageView key sizes to LineageView props

kwasi:
/area front-end
/area metadata
/assign @avdaredevil 

This is a followup change to #183 to make it easier to migrate `LineageView` to a standalone repo.  I'm working under the assumption that any file named `Lineage*.tsx?` will be moved to the Lineage module of the shared-components repo.

## Changes

* Moves px() helper function out of `Css.tsx` into `LineageCss.ts` to eliminate a non-lineage dependency.
* Make `cardWidth` and `edgeWidth` props that are passed to LineageView and propagated down to `LineageCard` and `EdgeCanvas` 
* Make current key sizes (cardWidth=260px, edgeWidth=120) default parameters that can be overridden by props passed to `<LineageView />`. The goal is to allow `<ArtifactDetails />` the page/application to drive the sizes of the shared component `<LineageView />` while providing reasonable defaults.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/metadata/185)

### DIFF
--- a/frontend/src/Css.tsx
+++ b/frontend/src/Css.tsx
@@ -374,5 +374,3 @@ export function _paddingInternal(units?: number, directions?: string): NestedCSS
 export function padding(units?: number, directions?: string): string {
   return style(_paddingInternal(units, directions));
 }
-
-export const px = (number: number): string => `${number}px`;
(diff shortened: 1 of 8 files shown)


kwasi:
This has no visual changes, but the changes are deployed to: https://kubeflow-ui-07.endpoints.kwasinti-kubeflow-dev.cloud.goog/_/metadata/

Issue #241, opened 2020-10-15
K8s Metadata watcher: Support Executions and Events


jlewi:
/kind feature

I think it would be great if users could do the following

1. Create a K8s resource (e.g. TFJob, PyTorchJab, KatibJob, K8s Job, Tekton TaskRun, Tekton PipelineRun, Argo, etc...)
1. Attach annotations to the resource indicating input and output artifacts
1. Apply the resource
1. See the lineage graph in KF metadata
    * Artifacts should show up as artifacts
    * The batch resource (TFJob, TaskRun etc....) should be stored in MLMD as an exeuction
    * Executions and artifacts should be connected via MLMD events to form the graph.

What we have today
    * [Watcher](https://github.com/kubeflow/metadata/tree/master/watcher) is a simple go program that watches the K8sAPIServer and logs resources as artifacts in MLMD

What we need todo
   * Modify the watcher to log specific types (K8s Job, TFJob, PyTorchJob, etc...) as Exeuctions (and not artifacts)
   * Modify the watcher to look for annotations defining input and output artifacts
   * Watcher creates MLMD events to link artifacts and annotations

@karlschriek , @aronchick @Swikar @Ark-kun  WDYT? Anyone have cycles to work on this?

Swikar:
I will look at it

Swikar:
Attach annotations to the resource indicating input and output artifacts. 

Can I have more detail or example related annotation used somewhere? 

I have already modified logging as execution rather than artifacts.

Thanks

jlewi:
> Can I have more detail or example related how to set annotation and artifact, have we used somewhere I can have a look?
https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

I imagine we would want to define some schema for annotations to contain input/output schemas. These might just be a list of the json representations of some of the MLMetadata schemas for artifacts.

> Do we want replace artifact with executions or keep both? I would assume replace with executions, So that input artifact annotation use for finding purpose - correct?

We want to keep both. The inputs and outputs should be stored as artifacts and then there should be an execution as well for the actual execution.

jlewi:
So here's a more precise example.

The dataset schema is currently defined here
https://github.com/kubeflow/metadata/blob/master/schema/alpha/artifacts/data_set.json

And an example is

```
{
            "annotations": {
                "mylabel": "l1",
                "tag": "data-set"
            },
            "apiversion": "v1",
            "category": "artifact",
[...]

PR #246, opened 2020-10-31
K8s watcher creates simple lineage graphs for K8s objects

jlewi:
* K8s objects will now be logged as executions instead of artifacts
* The watcher looks for annotations defining inputs and outputs
  * If the annotations are present they will be logged as artifacts
  * The watcher will then create events linking the inputs & outputs
    to the execution

* The net result is that we can generate lineage graphs just by attaching
  annotations to K8s object.

* The current implementation is pretty primitive; there are a number of
  issues to fix in follow ups

  * Handling multiple inputs and outputs
  * Handling the case where there already artifacts in the DB corresponding
    to the annotations

    * Right now a new artifact is created for each resource.

* Add an example job that we can use to test things.

Miscellaneous fixes

* For the watcher add Makefile rules to build with GCB
  * For the manifests create a kustomize file so its easy to overwrite
    the docker image with the latest image

* Check in the generated go client libraries for the metadata service.

* Update watcher deployment to v1.

* Change the resource_list.json to watch jobs not pods/services
  by default.

Related to: #241

/assign @karlschriek 
/assign @Swikar


Jeffwan:
I can help review the PR.

jlewi:
Test flakes should be fixed by #251

watcher/handlers/metalogger.go
```diff
@@ -122,18 +144,145 @@ func (l *MetaLogger) OnAdd(obj interface{}) error {
...
+		return []int64{}, nil
+	}
+
+	inputArtifact, err := toUnstructured(input)
+
+	if err != nil {
+		return []int64{}, err
+	}
+
+	gvk := inputArtifact.GetObjectKind().GroupVersionKind()
```
Jeffwan, line 216: This requires us to have GVK for the input. this makes us define more crds and build common wrapper on the inputs like `dataset`, `persist_model` etc. 

Do we want to support primitive type like a  string ?

watcher/handlers/metalogger.go
```diff
@@ -122,18 +144,145 @@ func (l *MetaLogger) OnAdd(obj interface{}) error {
...
+
+	mlmdTypeId, ok := l.gvkIds[gvk]
```
Jeffwan, line 218: If the server restarts, it will lose in memory mapping. maybe we can double check DB before we create it.

watcher/handlers/metalogger.go
```diff
@@ -122,18 +144,145 @@ func (l *MetaLogger) OnAdd(obj interface{}) error {
...
+
+		if err != nil {
+			log.Error(err, "Could not log artifact annotation", "annotation", annotation)
+			continue
+		}
+
+		// Create events connecting the execution and artifacts
+		for _, aId := range ids {
+			// TODO(jlewi): What should we set for the path if anythong?
+			e := &mlpb.Event{
```
Jeffwan, line 173: I remember Event has a path field but have not tried this before. Do you know if https://github.com/google/ml-metadata/blob/127cb87df7b3509140e321c401c77a9d0a8a223c/ml_metadata/proto/metadata_store.proto#L165-L174 this can be used here?

watcher/dockerfiles/resource_list.json
```diff
@@ -1,10 +1,7 @@
 [
     {
+        "Group": "batch",
```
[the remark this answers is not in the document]
jlewi: You can watch whatever K8s resources you want. Originally I think @zhenghuiwang selected Pods and Services for no particular reason.  In this PR i changed it to be a K8s job because i want to show you could submit a batch job and then collect metadata for it via annotations.

watcher/examples/dataset.yaml
```diff
@@ -0,0 +1,11 @@
+# Simple example of some YAML spec describing an artifact
+# This is intended to be attached as an annotation to the job.
+#
+# To convert this to json to attach as an annotation to the job
+# yq r ./examples/dataset.yaml --tojson
+apiVersion: metadata.kubeflow.org/v1alpha1
```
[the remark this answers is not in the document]
jlewi: This is the YAML that the metadata annotations attached to the Job are generated from. Its just more convenient as a human to read/write pretty formatted YAML as opposed to a single line of json.

watcher/handlers/metalogger.go
```diff
@@ -62,25 +79,26 @@ func NewMetaLogger(kfmdClient storepb.MetadataStoreServiceClient, gvk schema.Gro
 			// stores the metav1.Object
 			"object": mlpb.PropertyType_STRING,
```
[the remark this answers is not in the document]
jlewi: If the K8s object fits in etcd shouldn't it fit in the MetadataStore? I see the metadata store as a permanent record for all K8s "Jobs" so I would expect those resources to be able to fit.

watcher/handlers/metalogger.go
jlewi: It doesn't require CRDs. It requires that metadata schemas adhere to the Kubernetes resource model (https://docs.google.com/document/d/1RmHXdLhNbyOWPW_AtnnowaRfGejw-qlKQIuLKQWlwzs/edit#) which really just requires that they have metadata.

That seems like a good thing because that enables introspection just like we are doing here.

An example of a dataset is
https://github.com/kubeflow/metadata/blob/96285b257d999715a88befc659d722c895d7bbc5/watcher/examples/dataset.yaml

So we just need to include the fields "apiVersion and kind". That doesn't seem overly burdensome and offers a lot of nice properties.

watcher/handlers/metalogger.go
jlewi: I think that's ok. CreateArtifact is idempotent as long as the schema hasn't changed. We're just calling MLMD's PutArtifactTypeRequest method
https://github.com/kubeflow/metadata/pull/246/files/96285b257d999715a88befc659d722c895d7bbc5#diff-1b3125b98e9b2a5dba6c82b8938acbd4b1016241b32d7eb117b8abd3e8449852R271

watcher/handlers/metalogger.go
jlewi: I don't know what the path field is supposed to be used for.  I'm guessing its some way of distinguishing among the different inputs to an execution. e.g. if you have N values and the execution takes N inputs you need some way to match the values to input; I'm guessing that's what path is for.

watcher/examples/dataset.yaml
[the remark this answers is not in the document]
Jeffwan: Got it. It's a little confused at first glance. I thought this is some CRD.

watcher/handlers/metalogger.go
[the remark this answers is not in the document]
Jeffwan: I think it will fit, let's handle scalability issues later when we meet them

watcher/handlers/metalogger.go
Jeffwan: I misunderstand that. I thought `dataset` was crd but actually a definition used by manifest only. then it sounds good to me

Jeffwan (approved):
@jlewi https://github.com/kubeflow/metadata/pull/251 LGTM and please merge it and rebase upstream change.

jlewi:
I'm closing this because I no longer have merge rights and I'm not sure what's going on with metadata anymore.

Swikar:
Sure

